PR #5, opened 2015-07-08
Added a simulation of onBlur to enable final save events and the like

anateus:
The new `onBlur` event is documented. It also has the same semantics as the `onChange` handler: it expects the latest value, rather than the event object.

This is intended to enable not needing to have more heavy-weight save events on every change, but rather saving it for when a user is done providing input.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -22,6 +22,7 @@ var Datetime = React.createClass({
 	},
 	propTypes: {
 		date: TYPES.object,
+		onBlur: TYPES.func,
 		onChange: TYPES.func,
 		locale: TYPES.string,
 		input: TYPES.bool,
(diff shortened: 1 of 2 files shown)


arqex:
@anateus Thanks!

You know that react-datetime is still in process to be stable. It would be great to have know your thoughts about the API.

From my point of view I have some doubts about the way of making it work as a datepicker/timepicker only. Currently if you pass prop `dateFormat` and not a `timeFormat` you only get the datepicker. The same happens in the other direction. But this way it is not possible to use the locale default format, so I was thinking about change it.

Also I was thinking about deprecating `minDate` and `maxDate` and having a prop `validDate` that accepts a function to let the developer pick fine grained valid dates.

What do you think? Do you have other ideas or needs?

Issue #6, opened 2015-07-09
Webpack Error


aleemb:
Thanks for the wonderful lib, it's has the lightest footprint of the other libs I have explored.

I ran into the following error while trying to include it in my project:
```
~> npm install react-datetime --save
~> webpack --watch
# var Datetime = require('react-datetime');

ERROR in ./js/test.jsx
Module not found: Error: Cannot resolve module 'react-datetime' 
```

PR #7, opened 2015-07-09
Fix module entry point #6

aleemb:


arqex:
Thanks for the PR!

I am asking for help with the API:
https://github.com/arqex/react-datetime/issues/8

It would be great to have your thoughts.

Cheers

PR #9, opened 2015-07-09
IE support and custom classNames

anateus:
* This allows passing in class names for the input form (this allows me to reuse global styles for input fields that I have).
* It replaces using `dataset` with an explicit attribute request, which is compatible with all browsers even if it's slightly less pretty. If you were doing more dynamic dataset access this would be even less pretty, but thankfully you're only using one key, so hopefully you're OK with this.
* It adds another dependency to support `classlist`. The polyfill is pretty tiny, if you'd prefer to keep dependencies down, I can integrate the polyfill directly.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -21,6 +21,7 @@ var Datetime = React.createClass({
 		time: TimeView
 	},
 	propTypes: {
+		className: TYPES.string,
 		date: TYPES.object,
 		onBlur: TYPES.func,
 		onChange: TYPES.func,
@@ -36,6 +37,7 @@ var Datetime = React.createClass({
 	getDefaultProps: function() {
 
 		return {
+			className: 'form-control',
 			date: new Date(),
 			viewMode: 'days',
 			inputProps: {},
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "react-datetime",
-  "version": "0.4.0",
+  "version": "0.4.3",
```
anateus: I think this is only a diff because I rebased on top of the upstream master, so this is a change from the previous commit, even though it's not a change from your master... Weird...

anateus:
Bumping this. Is there something blocking this from being merged, or you've just not gotten around to it? Totally fine if it's the latter, would just like to know :)

DateTime.js
```diff
@@ -275,7 +279,7 @@ var Datetime = React.createClass({
 			children = [ DOM.input( assign({
 				key: 'i',
 				type:'text',
-				className:'form-control',
+				className: this.props.className,
```
arqex: Hi @anateus 

Sorry about my late reply, I've been very busy :)

No problem about the polyfill or not using the dataset, but there is already a way of setting the className for the input using the prop `inputProps`. 
```js
<Datetime inputProps={ {className: 'myAwesomeClass' } } />
```
I would reserve the `className` prop for the component wrapper instead. If you modify this I will merge the PR.

Thanks for your collaboration, it is really worthy :)

anateus:
Updated, reverting the classNames stuff. Do you know when you expect to have the onChange stuff you mentioned on your other PR ready? I'm currently integrating this into a project, and want to know when to expect this to land on NPM, or if I should roll my own package while I wait. Appreciate your quick turnaround!

PR #10, opened 2015-07-09
onChange had non-existant variable

anateus:
This stacks on top of my previous PR. If you don't accept that one, I can either close and re-open a 2 commit PR, or you can just cherry-pick the relevant commits.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -115,11 +115,13 @@ var Datetime = React.createClass({
 			localMoment = this.localMoment( value )
 		;
 
+
 		if (localMoment.isValid()) {
-			this.setState({
+			return this.setState({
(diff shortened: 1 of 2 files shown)


arqex:
Yeey @anateus 

Thanks for this, but I have prepared some updates for onChange an onBlur and this change will be discarded. So if you make modifications, please do them into the other PR.

The main change coming is `onChange` and `onBlur` will return a date object (moment.js) to be consistent with the `date` prop, making input/output consistent.

Cheers,
Javi

anateus:
OK, so that means the two bugs fixed here, will be fixed in your changes as well? In that case, I'll close this PR.

Issue #11, opened 2015-07-22
Buttons tags does not have type=button


yanickouellet:
All `<button>` tags do not specify a `type` attribute, which cause the page to be submitted if the picker is in a `form` tag and we change month using the arrows. Here is an exemple: http://codepen.io/anon/pen/EjeOJZ Using `<button type="button">` would fix the issue, since `button` without type are considered as submit by default. 

I can make a pull request to fix that issue in the following days.

arqex:
Sure, I'll be glad to merge your PR.
Thanks very much

yanickouellet:
The PR has been merged, I think this issue can be closed.

PR #12, opened 2015-07-23
Add type="button" to buttons to prevent page submit

yanickouellet:
The issue is described in #11


arqex:
Thanks for your PR @yanickouellet 

This is going to be in the next version of react-datetime. The release candidate.

PR #25, opened 2015-10-13
Add a 'strictParsing' prop which uses Moment's strict parsing.

nbrendler:
Hey, I'm using your component and would like to be able to use Moment's strict parsing, so if a user types input like
```
2015-10-12xyz
```
Moment detects this as invalid, or in the case of this component, it returns the string `2015-10-12xyz` in the onChange handler rather than a Moment object. Without strict parsing, the parser is a lot more lenient and will happily return a moment representing `2015-10-12` for this input.

This PR adds the functionality as a toggleable prop. Let me know what you think.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -33,7 +33,8 @@ var Datetime = React.createClass({
 		viewMode: TYPES.oneOf(['years', 'months', 'days', 'time']),
 		isValidDate: TYPES.func,
 		minDate: TYPES.object,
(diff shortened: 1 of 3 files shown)


arqex:
I think it is a great idea. I will merge it tomorrow and I think I will set the default value to true. I don't think anybody to like not strict parsing.

Thanks.

nbrendler:
Nice! strict=true works for me as the default.

PR #30, opened 2015-10-24
Add instructions about using within a form

willm:
It took me a while how to actually get the data back out of the widget, so I added some instructions in hope that it will help someone out in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,19 @@ render: function() {
 ```
 [See this example working](http://codepen.io/arqex/pen/BNRNBw).
 
+To use within a from, add a name property.
+
+```
+render: function() {
+  var inputProps = {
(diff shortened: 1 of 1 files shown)


simeg:
Hey @willm and thanks for your contribution! I think your changes are well needed, I'll try to add them to the README when I find the time. Thanks.

simeg:
I added a small clarification to the README in this commit: 2d2690d422075c8886c1f1d5d53e1586cb36b85c.

I'm currently working on more examples where I will have one example with a form, hopefully this will be sufficient for people wanting to use the component with a form. 

I don't want to clobber the README with example use-cases like these. The information is valuable but I think it should be stored with other examples.

willm:
Great thanks some :)

PR #39, opened 2015-11-22
Add onFocus callback prop

christopherscott:
I found it helpful to get a callback when the widget itself gains "focus". My use case is where I use two datetime pickers in tandem, and would like to be able to tab between them like normal inputs.


christopherscott:
just realized this was already handled by: #27 - closing :)

PR #49, opened 2015-12-28
External react-dom

imothee:
React-dom is required but not externalized. This causes the dist file to be huge and adds duplicate dom warnings from React when included.

I have updated the config and rebuild the dist files.

Some minor things to note but not fixed - would you like another pr?

* "gulp.cmd" no longer exists when doing npm install, the only bin file is "gulp"
* the webpack config is not used by gulp meaning externals have to be updated in multiple places


imothee:
Closing since #63 was merged, should close #50 as well. Three requests with the exact same fix  :)

imothee:
That said, I'd be keen on feedback around the gulp issues I raised.

PR #50, opened 2016-01-03
Make ReactDOM an external dependency; allow compiling with Webpack

nbudin:
I've made a couple build improvements while integrating react-datetime into my own app and thought I should make them available if you're interested.

First, I've made ReactDOM an external dependency just like React itself is.  This change reduces the compiled size of react-datetime.js from over 700KB to about 30KB.  This means the user will have to separately include ReactDOM in their app, but chances are they are already using some other component that needs it, so this shouldn't be a big deal in most cases.

Second, I've updated the webpack.config.js to allow react-datetime to be compiled using `webpack --config webpack.config.js` on the command line, and added a source map generation option.  This made it much easier for me to debug an issue I was having in my app (which turned out to be a bug in my code and not react-datetime, thankfully!).


imothee:
Bumping this or my original pr #49

nbudin:
I see that #63 was merged, so let's close this.

PR #52, opened 2016-01-04
Handling tabbing out of the date picker

gcedo:
Hi!

The goal of this PR is to automatically close the DatePicker when the user tabs out of the component. 

I used `onKeyDown` event binding rather than `onBlur` because otherwise the DatePicker would have closed also when back-tabbing from the `Next` arrow button to the input field.

Cheers,
Edo

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -129,6 +129,12 @@ var Datetime = React.createClass({
 		this.setState( update );
 	},
 
+	handleOnKeyDown: function( e ) {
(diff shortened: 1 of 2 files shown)


simeg:
Hi @gcedo. Sorry for the very late response.

It looks like this functionality was added a while back in this commit: 9012e8199a9db78b4a9d381378cf9843ec8c1a1e

Thanks for your contribution, but as this already exists in the component I will close this PR.

PR #54, opened 2016-01-04
Added support for HTML required attribute.

gcedo:
Hi!

The goal of this PR is to support the HTML `required` attribute, which may be useful when the DatePicker is used in a form.

In order to achieve this, I have added a `required` prop, which is simply passed down to the input field.

Cheers,
Edo


arqex:
Hi @gcedo 

Thanks for the PR, but the required attribute can be set using the `inputProps`.

```js
<DateTime inputProps={{required:true}} />
```
http://codepen.io/arqex/pen/wGBKBB

PR #55, opened 2016-01-07
Reset currentView after clickoutside

lizheming:
This behavior is the default in http://eonasdan.github.io/bootstrap-datetimepicker/ and I think that's friendly for using.


light049:
I use this change but it could not work as I expected.
I change it as below:
`this.setState({
	open: false,
	inputValue: this.state.selectDate ? this.state.inputValue : this.props.value
});
`

lizheming:
@light049 Thank you for your response. You may misunderstand what I mean. I set `viewMode="days"` for react-datetime, and I wish calendar will go back the view mode after I go into month or year view mode when I click outside. So I add code to reset view mode when click outside.

light049:
@lizheming  Haha~ yes, you are right. I'm sorry I missed that... ^^"
You add code to set the view to be the initial view mode.
What I want to do is to reset the displayed input value to be a valid datetime string(the previous valid datetime) when click outside.

Thanks for your reply~ 👍

simeg:
Hi @lizheming, thanks for your contribution.

To me it looks like this change will fallback to `currentView: time` if `viewMode` is not set. Is that wanted behaviour, and is the bootstrap version really doing that? I'm not sure if I understand.

simeg:
Closing due to lack of activity.

PR #63, opened 2016-02-18
Add react-dom as external dependency

poke:
Starting with d359ebdba9, React was included in the dist builds. This does not only make the file size explode but also causes conflicts when you try to use the dist build on a page that already has loaded another React version.

The cause for this was that `react-dom` was required in `src/onClickOutside.js` but it was not added as an external dependency. So webpack would resolve all the references itself (going from `react-dom` to `react/lib/ReactDOM` and from there into all the individual small modules), and eventually add the whole React framework to the output.

This pull request adds `react-dom` as an external dependency, so it’s *not* included in the build, making the dist version small again.


arqex:
Thanks for the PR, and sorry for not replying before. Now youcanbook.me is sponsoring react-datetime, so updates will come more frequently :)

PR #64, opened 2016-02-18
Fix bug where rdtActive not getting set

munik:
We noticed that rdtActive was not getting set at all. That's because isSame is being called incorrectly for Moment v2+.

The equivalent functionality is to pass a moment as the first argument and the string "day" as the second argument:

"When including a second parameter, it will match all units equal or larger. Passing in month will check month and year. Passing in day will check day, month, and year."
http://momentjs.com/docs/#/query/is-same/


arqex:
Thanks for the PR, and sorry for not replying before. Now youcanbook.me is sponsoring react-datetime, so updates will come more frequently :)

PR #67, opened 2016-03-01
Months selectors is empty if not english locale.

Megabyteceer:
I tried to use Your datapicker on my page, where moment is swiched to russian location.
When i click year line, all month choosers was empty. This fix for my problem, maybe its will be helpfull for you.

Thanks for datapicker.

### DIFF
--- a/src/MonthsView.js
+++ b/src/MonthsView.js
@@ -54,7 +54,7 @@ var DateTimePickerMonths = React.createClass({
 	},
 
 	renderMonth: function( props, month, year, selectedDate ) {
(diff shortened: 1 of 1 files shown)


AlexKovalevych:
I've faced with the same issue, but your solution showed me english month names instead of russian, please check this alternate PR i made: #76 .

arqex:
@Megabyteceer & @AlexKovalevych thanks so much for your contributions. I have merged

https://github.com/YouCanBookMe/react-datetime/pull/95

So I hope this problem is already fixed in v.2.3.3

PR #68, opened 2016-03-14
added isDisabled prop to datetime field

export-mike:
allowing the text fields to be disabled


tmartensen:
Bumping, could this get merged?

export-mike:
Umm this was so long ago I'll have to take a look next week and cast my mind back to its use case

light049:
Will this prop be covered with inputProps={{disabled: true}} ?

simeg:
Hi @export-mike and thanks for the PR, 

As @light049 mentions you can use the `inputProps` prop to disable the input field, just like this:

`<Datetime inputProps={{ disabled: true }} />`

PR #70, opened 2016-03-19
Feature / DayView footer button to set/reset date to now

tomasztunik:
Added ability to add reset to now button with visiblity based on presence of nowLabel prop

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -36,7 +36,8 @@ var Datetime = React.createClass({
 		isValidDate: TYPES.func,
 		open: TYPES.bool,
 		strictParsing: TYPES.bool,
-		nowLabel: TYPES.string
+		nowLabel: TYPES.string,
+		setLabel: TYPES.string
 	},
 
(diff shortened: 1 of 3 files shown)


arqex:
Hey @tomasztunik 

Thanks for this pull request. I think the reset button is a powerful idea, but I am not sure if this could be better done letting the developers customizing the datepicker code:

https://github.com/YouCanBookMe/react-datetime/pull/82

I think that some users would like to reset to any other date different from now, or having a custom behaviour for that reset button, and adding the button using a prop is too strict for these cases.

Still thinking about what would be the best solution.

tomasztunik:
More generic approach might be better indeed. Especially if it was possible to control it per view (year, month, date, time etc, or in a more general way to embed something in all the footers at once). Feel free to close this pull request and I'll keep the fork live in case anybody needs something similar or inspiration how to get it done as we use it for one of our projects anyway.

simeg:
I'm closing this. Continue discussing in #153.

PR #73, opened 2016-04-05
Month or Year only date picker

sridharAJ:
Hi,

I have a use case were user will select only month or year from datetime picker .

if month only - we should not show day selection. 
if year only - we should not show month selection. 

For more info  : [refer here](https://eonasdan.github.io/bootstrap-datetimepicker/#min-view-mode)

Thanks.


arqex:
Thanks for the PR,

I'll merge it, but I think it still need some work, there are some date formats that are not handled by this PR, also the input's format in months and year mode shouldn't show the time.

I'll work on them and merge.

Cheers

sridharAJ:
Hi,

Thanks for you reply.

Please try to  merge it before this weekend as coming Monday our
application is scheduled to go production.

I will be grateful if you could do it.


Regards,
Sridhar Arumugam.

PR #76, opened 2016-04-11
fix month names for locales with standalone/format options

AlexKovalevych:


fend25:
according to https://travis-ci.org/YouCanBookMe/react-datetime/builds/122182392 test have failed because of wrong test (test couldn't find a module).

please fix the test (or travisCI config) and merge the PR - with locale months doesn't render.

arqex:
https://github.com/YouCanBookMe/react-datetime/pull/95

Issue #41, opened 2015-11-25
Customizing the time picker


rchanou:
I see how to do formatting in the input field, but is there any way to customize the time picker? I want it to only go from 1-12 for the hour picker and add an AM/PM toggle button.

0xF013:
+1

rchanou:
react-bootstrap-datetimepicker has the opposite problem (only AM/PM, no 24-hour timepicker), though it looks like they're working on releasing a fix: https://github.com/quri/react-bootstrap-datetimepicker/issues/105

0xF013:
if only that thing had seconds picker

arqex:
We probably will be able to add this feature defining a `timeFormat` prop with an `a` or `A`

http://momentjs.com/docs/#/displaying/

so making `<DateTime timeFormat={hh:mm A} />` would switch to the AM/PM mode.

crobinson42:
+1 
I'd like to see this feature as well. I'm glad to help with this enhancement with contribution guidelines.

bandanaman:
Switching to AM/PM mode would be great feature, but since widget and date format are two separate things, maybe better idea is to specify time type with an attribute: `< DateTime twelveHourClock={ false } />`

andrewmclagan:
plus one, would be a great feature

bandanaman:
I added this functionality in my fork https://github.com/bandanaman/react-datetime and created pull request. It works as arqex described in his post - searches for " A" in entered time format.

andrewmclagan:
Will take a look. Thanks!

ecerroni:
+1 
I'd like to see this feature as well.

andrewmclagan:
Any chance this PR will be merged?

simeg:
Closing this as referenced PR is merged.

rchanou:
Awesome, will the latest version be pushed to npm soon?

PR #91, opened 2016-06-16
Customizing the time picker

bandanaman:
I extended time view widget to switch between am/pm and 24h clock modes basing on time format, as arqex described in his post https://github.com/YouCanBookMe/react-datetime/issues/41#issuecomment-190843119

### DIFF
--- a/src/TimeView.js
+++ b/src/TimeView.js
@@ -23,20 +23,33 @@ var DateTimePickerTime = React.createClass({
 			}
 		}
 
+		let daypart = false;
+		if( this.props.timeFormat.indexOf(' A') != -1  && this.state != null ){
+			daypart = ( this.state.hours > 12 ) ? 'PM' : 'AM';
+		}
+
 		return {
 			hours: date.format('H'),
 			minutes: date.format('mm'),
 			seconds: date.format('ss'),
 			milliseconds: date.format('SSS'),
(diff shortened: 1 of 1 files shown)


serkandurusoy:
Wow this is exactly what I also needed! Thanks. @YouCanBookMe do you think you can get this merged and published?

kaiyes:
@arqex this would help me as well. it would be really great if this is merged !

ecerroni:
I'd like to see this PR accepted and merged. I also need that feature ASAP.

janoschpizza:
+1 on this

robfallows:
+1 here too. Would be really useful!

arqex:
Hey @bandanaman 

Thanks for the PR, I think it will need more work to also accept the `a` parameter in the timeFormat, 

maybe checking `timeFormat.match(/a$/i)` do the trick. Since a lot of people needs this feature I decided to merge it now.

serkandurusoy:
Thank you all, this feature will be very valuable for me!

kaiyes:
Thanks a lot !

kahurangitama:
I have an app where in some places I need to use regular date picker (without time picker) and want to use the same plugin but without this feature. Is there some option to turn off the ability of time picking?

PR #95, opened 2016-07-01
Fixed displaying of months for 'RU' and other locales

g-gray:


arqex:
@g-gray Thanks so much for your contribution, and sorry for the delay.

PR #98, opened 2016-07-01
Update README, build scripts, example and indentation

simeg:
So I wasn't able to get all four changes in separate pull requests in a clean way - sorry about that.

About the updated README: I changed the example on CodePen so that the input element now has a black border so it's easier to see it (me and a colleague both thought that the component was broken when we opened the CodePen as we didn't see the white input field at first).

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ This project started as a fork of https://github.com/quri/react-bootstrap-dateti
 Usage
 ===============================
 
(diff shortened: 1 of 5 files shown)


simeg:
@arqex @chollier ping

arqex:
@simeg Thanks so mucho for your contribution, and sorry for the delay.

PR #100, opened 2016-07-08
Add timeConstraints prop to set custom min, max and step values

loris:
It adds a `timeConstraints` prop which allow customising `min`, `max`, `step` constraints for hours/minutes/seconds/milliseconds. Here is an example:
```
<DateTime
  timeConstraints={{ hours: { min: 8, max: 22 }, minutes: { step: 10 } }}
/>
```
It renders a DateTime picker where you can only select hours from 8 to 22, and minutes increment by 10 instead of 1.


arqex:
@loris 

Thanks for this idea, but I would probably unify all the properties in one. Probably something like `timeConstraints`, an object that wraps all yours `hours`, `minutes`... props.

I am not very sure about the `pad` option, if you see [moment.js docs](http://momentjs.com/docs/#/displaying/) you can pad your time parts using the `timeFormat` prop.

loris:
Hi @arqex,
Just updated my PR with your suggestions (unified props into `timeConstraints`, added it to propTypes, removed the `pad` option), added a quick test case. By the way, I added `moment` and `react-dom` to `devDependencies` in order to `npm test`

arqex:
Thanks @loris 

I'll try to review the PR tomorrow, I am really busy today. But I appreciate much the work :)

tmartensen:
Sorry to butt in, but I checked out your branch, @loris, but if you set a step of 15,  it looks like the up arrows work correctly, but the down arrow starts at 59, then goes to 44, etc.

loris:
@tmartensen Sorry about that, I overlooked the "looping" behavior (going over the max or below the min values) as it was not possible in my own use case. I updated the PR with a fix.

tmartensen:
@loris No apologies to me are necessary. Thanks for implementing this! It's a feature that I definitely need.

arqex:
Thanks @loris for your effort on this. Really appreciated!

Issue #108, opened 2016-07-19
Run linter as a pre-push hook


simeg:
It would be nice if the linter would run automatically before every push and fail to push if the linter finds anything wrong. This would increase the quality and readability of the code written as less bugs would be introduced and the syntax would be more consistent.

It could also be argued that it should be used as a pre-commit hook.

PR #109, opened 2016-07-19
Add pre-commit hook

simeg:
Referencing issue #108.

I decided for a pre-commit over a pre-push, as I think it's better to see your errors while performing your commits instead of a bunch of problems when pushing. 

The pre-commit hook will run the `lint` and the `test` tasks. If any of them fail the commit will not be accepted. You can disable the hook for the current commit by using `--no-verify` on your commit, e.g. `git commit -m 'Commit message' --no-verify`. This should only be used for commits that break any test but is fixed with a later commit in the same push (if that makes sense :) ).

**Be sure to run `npm install` to get the [`pre-commit`](https://github.com/observing/pre-commit) package.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,12 +142,6 @@ Contributions
 ===============================
 Any help is always welcome :)
 
-**Please use the linter before submitting your pull request.**
-
-```
-npm run lint
-```
-
 ### [Changelog](CHANGELOG.md)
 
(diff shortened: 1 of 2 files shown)


arqex:
Thanks @simeg, I couldn't merge it today, next week I'll have a look at it.

PR #110, opened 2016-07-19
Add test task with watch option

simeg:
Writing tests often requires them to be run multiple times before they pass. It can be a hassle switching focus back and forth between your editor/IDE (to write code) and terminal (to init test task). By running this npm task the mocha instance will watch for changes in files and automatically re-run the test task, saving you time and energy.

It is run like `npm run test:watch`.


arqex:
@simeg Really nice this feature. Thanks!

PR #114, opened 2016-07-30
Fix bugs with the time picker

jackdeadman:
Fixed a couple issues with the time picker. One being that the timer would go to 0:00AM rather than 12:00AM. The other being I fixed an issue with the time picker not displaying the AM when first loaded.


tmartensen:
Bumping, could this get merged?

simeg:
Hey @jackdeadman, nice findings. A few issues though, when I run this code locally it doesn't actually resolve the bugs you found (which indeed are bugs). And secondly, the code does cause the test `dateFormat` to render the warning `Warning: Each child in an array or iterator should have a unique "key" prop.`. 

The code base might have changed a bit since you first wrote this, which would be unfortunate. If you want you can try to make a new PR with code that works. In that case you could also get rid of the warning, because I don't want to merge this with that warning included. It may work with the warning but it may be the root of a bug in the future. 

Let me know what you think. Thanks!

jackdeadman:
Interesting it seems to be working for me, it must be the changes made since I wrote it. I'll update it, sorry about that.

My version behaves likes this...
4obr12wjym

simeg:
Can you please rebase with master to see if you get a different result? In that case I'm doing something wrong.

jackdeadman:
I just tested the component with the latest version. The warning isn't caused by my code, it's in the repo already. My code seems to be working with the latest version, I'm probably doing something silly, sorry this is my first pull request.

Your current version
original

My version
new

jackdeadman:
The issue with the warning was caused by the AM/PM picker, I've just pushed a fix for that as well. This should work :)

simeg:
@jackdeadman thanks for resolving that bug. As you see your commit history is a bit messed up, I'm not sure how to resolve it in a clean way. My suggestion is to 1) clone this repo again 2) make your three commits again and 3) create a new PR.

src/TimeView.js
```diff
@@ -56,7 +56,7 @@ var DateTimePickerTime = React.createClass({
 		return '';
 	},
 	renderDayPart: function() {
-		return DOM.div({ className: 'rdtCounter'}, [
+		return DOM.div({ className: 'rdtCounter', key: 'c'}, [
```
simeg: What is the value `c` referring to?

jackdeadman: It just had to be unique, I saw "c" was used elsewhere in the code, but after checking the code further, "dayPart" would be a better name for that key.

jackdeadman:
Alright, I'll close this PR and try again. Sorry about this

Issue #116, opened 2016-08-04
Multiple date pickers with different time constraints interfere


steos:
When I have two or more time pickers with different time constraints they will interfere with each other.

Check out this minimal example: http://codepen.io/anon/pen/wWYzxp?editors=1010

The first picker has a step constraint of 15 for minutes, the second picker has none.
If you use the first picker and then the second one it will suddenly use the timeConstraints of the first one.

This even happens when the timeConstraints are set explicitly on both pickers.

PR #117, opened 2016-08-04
Create time constraints object on mount instead of assigning to it

andreasgrill:
This should resolve Issue #116.


simeg:
Thanks @andreasgrill. This is really useful!

PR #82, opened 2016-05-12
added option to add child components to DateTime

besuikerd:
as per request in #45 and #74, added an option to 'decorate' the input component. 

usage:

``` javascript
<DateTime 
  inputDecorator={
    (input) => <div>
      <div className="someDecoration">
        {input}
      </div>
    </div>
  }
/>
```

PR #134, opened 2016-09-07
Add label prop to have a label text in input

light049:
I just added a label prop for user who want the input to have a label text at left side.
And update required layout followed bs styles.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -28,7 +28,10 @@ var Datetime = React.createClass({
 		onChange: TYPES.func,
 		locale: TYPES.string,
 		input: TYPES.bool,
-		label: TYPES.string,
+		label: TYPES.oneOfType([
+			TYPES.string,
+			TYPES.element
+		]),
 		// dateFormat: TYPES.string | TYPES.bool,
 		// timeFormat: TYPES.string | TYPES.bool,
 		inputProps: TYPES.object,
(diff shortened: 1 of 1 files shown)


simeg:
Hi @light049, thanks for the PR.

I'm not sure how I feel about letting this component have the responsibility for the label. The label *does* belong to the form and the SEO would increase (if the `<label>` element was used). 

What do you think @arqex?

light049:
If that's naming issue, might just use other prop name like inputLabel?

simeg:
The naming is not the problem, the problem is defining what the component should be responsible for. Let's wait and see what @arqex thinks, as he is the author for the component :) If he thinks it's a good idea I think we can re-write the code a bit before we merge it, but we can discuss that later.

simeg:
Thinking about this, the **only** reason I see to merge something like this is that the SEO score would increase if you wanted to use a label. Otherwise I think it's nice to keep this component clean of any additional stuff. You can always add some text that will act as a label outside of the component. 

This is of course just my idea of what the component should and should not do. Any good reasons why you think this should be included in the component @light049?

arqex:
From my point of view, we should create a way of completely edit the markup of the component. There is a lot of people who are requesting this kind of changes like:

https://github.com/YouCanBookMe/react-datetime/pull/82
https://github.com/YouCanBookMe/react-datetime/pull/152

Somebody need a reset button, other a label, other an icon in the input...

I don't think react-datetime needs to include any code to please those specific users, but a way of letting them customize the component as they need. Currently, we have the `renderMonth`(day, year) props. Maybe we need to add the `renderInput` and `renderView` ones.

simeg:
@arqex Good! Then we share the same opinion, because I was literally just writing issue #153, which would resolve this and many other issues.

simeg:
I'm closing this. Continue discussing in #153.

PR #138, opened 2016-09-10
Fixing argument type for onBlur on onFocus events

skyflyer:
Since I'm getting errors (from webpack):

```
ERROR in .../react-datetime/react-datetime.d.ts
(57,16): error TS7006: Parameter 'e' implicitly has an 'any' type.

ERROR in .../react-datetime/react-datetime.d.ts
(63,16): error TS7006: Parameter 'e' implicitly has an 'any' type.
```

I though I could fix it.

### DIFF
--- a/react-datetime.d.ts
+++ b/react-datetime.d.ts
@@ -54,13 +54,13 @@ declare module ReactDatetime {
     /*
      Callback trigger for when the user opens the datepicker.
      */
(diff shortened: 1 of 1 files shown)


simeg:
Thanks @skyflyer! I'm not using the TypeScript version of the component so PRs like these are highly appreciated.

PR #139, opened 2016-09-11
Add CONTRIBUTION file

simeg:
This PR adds a CONTRIBUTION.md file which will contain guidelines for making a contribution to this repository. 

The idea with this file is not to enforce rules but to give guidelines. In a perfect world I think everyone contributing should follow these guidelines for a shared way-of-working, but I don't think we should deny PRs because they are not following the guidelines.

I hope these guidelines will have people reflect over their current workflow and think about why they are doing things the way they are, because it's easy to get stuck in a workflow where the focus is on producing code and nothing else. When multiple people are involved in a project it's important to make things understandable for everyone, not just yourself. 

With that said I hope to get some feedback on these guidelines. 

**Note**, I'm not just looking for feedback from maintainers of this project - I'm interested in feedback from anyone :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+# Contributing
+
+:raised_hands::tada: First off, thanks for taking the time to contribute! :tada::raised_hands:
+
+The following is a set of guidelines for contributing to react-datetime. The purpose of these guidelines is to maintain a high quality of code *and* traceability. Please respect these guidelines.
+
+## General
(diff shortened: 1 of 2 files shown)


arqex:
Nice! the best part of it are the emojis at the begining!! 💃 

Was joking, it looks great, there was time of having the contribution file!

simeg:
@arqex cool! So if you don't have any feedback, let's wait a few days for others to have time to give feedback and then merge.

layneanderson:
I think the guidelines are great and explains the purpose and methods very well.

:+1:

PR #146, opened 2016-09-15
Updating typings config to reflect actual callback parameters

skyflyer:
## Description

This is a typescript typings file with corrected definitions for properties and callbacks on `Datetime` object.

## Motivation and Context

This PR should be merged to avoid Typescript compilation errors that result in using the callbacks with wrong callback parameters.

## Checklist

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] All new and existing tests passed.
- [X] No javascript code has been changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,9 +55,9 @@ API
 | **input** | boolean | true | Whether to show an input field to edit the date manually. |
 | **open** | boolean | null | Whether to open or close the picker. If not set react-datetime will open the datepicker on input focus and close it on click outside. |
 | **locale** | string | null | Manually set the locale for the react-datetime instance. Moment.js locale needs to be loaded to be used, see [i18n docs](#i18n).
-| **onChange** | function | empty function | Callback trigger when the date changes. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If it isn't, the value of the input (a string) is returned. |
+| **onChange** | function | empty function | Callback trigger when the date changes. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If the date in the input is not valid, the callback receives the value of the input (a string). |
 | **onFocus** | function | empty function | Callback trigger for when the user opens the datepicker. |
-| **onBlur** | function | empty function | Callback trigger for when the user clicks outside of the input, simulating a regular onBlur. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If it isn't, the value of the input (a string) is returned. |
+| **onBlur** | function | empty function | Callback trigger for when the user clicks outside of the input, simulating a regular onBlur. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If the date in the input is not valid, the callback receives the value of the input (a string). |
 | **viewMode** | string or number | 'days' | The default view to display when the picker is shown. ('years', 'months', 'days', 'time') |
 | **className** | string or string array | `""` | Extra class name for the outermost markup element. |
 | **inputProps** | object | undefined | Defines additional attributes for the input element of the component. |
--- a/README.md
+++ b/README.md
@@ -42,9 +42,9 @@ render: function() {
 | **input** | `boolean` | `true` | Whether to show an input field to edit the date manually. |
 | **open** | `boolean` | `null` | Whether to open or close the picker. If not set react-datetime will open the datepicker on input focus and close it on click outside. |
 | **locale** | `string` | `null` | Manually set the locale for the react-datetime instance. Moment.js locale needs to be loaded to be used, see [i18n docs](#i18n).
-| **onChange** | function | empty function | Callback trigger when the date changes. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If the date in the input is not valid, the callback receives the value of the input (a string). |
-| **onFocus** | function | empty function | Callback trigger for when the user opens the datepicker. |
-| **onBlur** | function | empty function | Callback trigger for when the user clicks outside of the input, simulating a regular onBlur. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If the date in the input is not valid, the callback returned. |
+| **onChange** | `function` | empty function | Callback trigger when the date changes. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If the date in the input is not valid, the callback receives the value of the input (a string). |
+| **onFocus** | `function` | empty function | Callback trigger for when the user opens the datepicker. |
+| **onBlur** | `function` | empty function | Callback trigger for when the user clicks outside of the input, simulating a regular onBlur. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If the date in the input is not valid, the callback returned. |
 | **viewMode** | `string` or `number` | `'days'` | The default view to display when the picker is shown (`'years'`, `'months'`, `'days'`, `'time'`). |
 | **className** | `string` or `string array` | `''` | Extra class name for the outermost markup element. |
 | **inputProps** | `object` | `undefined` | Defines additional attributes for the input element of the component. For example: `placeholder`, `disabled` and `required`. |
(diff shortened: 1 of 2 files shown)


react-datetime.d.ts
```diff
@@ -50,17 +50,17 @@ declare module ReactDatetime {
      only parameter, if the date in the input is valid. If it isn't, the value
      of the input (a string) is returned.
      */
-    onChange?:(x: string) => void;
+    onChange?:(x: string|any) => void;
```
simeg: Please add space between `onChange?:` and `(x: string ...)` to comply with styling of the rest of the file.

react-datetime.d.ts
```diff
@@ -50,17 +50,17 @@ declare module ReactDatetime {
...
      */
-    onFocus?: (e : FocusEvent) => void;
+    onFocus?: () => void;
     /*
      Callback trigger for when the user clicks outside of the input, simulating a regular onBlur.
      The callback receives the selected moment object as only parameter, if the date in the
      input is valid. If it isn't, the value of the input (a string) is returned.
      */
-    onBlurs?: (e : FocusEvent) => void;
+    onBlur?: (e : string|any) => void;
```
simeg: Nice find with the typo!

react-datetime.d.ts
simeg: Doesn't `onChange` accept a `function` and not `string` or `any` as input?

react-datetime.d.ts
```diff
@@ -50,17 +50,17 @@ declare module ReactDatetime {
      only parameter, if the date in the input is valid. If it isn't, the value
      of the input (a string) is returned.
      */
-    onChange?:(x: string) => void;
+    onChange?:(x: string|any) => void;
     /*
      Callback trigger for when the user opens the datepicker.
      */
-    onFocus?: (e : FocusEvent) => void;
+    onFocus?: () => void;
```
simeg: Doesn't `onFocus` accept a `function` and not `string` or `any` as input?

react-datetime.d.ts
simeg: Doesn't `onBlur` accept a `function` and not `string` or `any` as input?

react-datetime.d.ts
simeg: Doesn't this actually return something according to the docs? 

> The callback receives the selected moment object as only parameter, if the date in the input is valid. If it isn't, the value of the input (a string) is returned.

Or is the documentation a bit confusing? First it mentions the parameter, and then it says "If it isn't - return string". Maybe I'm just misunderstanding something here?

react-datetime.d.ts
simeg: Same thing about `onBlur` as I mentioned above with `onChange`. I'm a bit confused, I'd love an explanation if you have one :)

react-datetime.d.ts
```diff
@@ -77,27 +77,27 @@ declare module ReactDatetime {
      Define the dates that can be selected. The function receives (currentDate, selectedDate)
      and should return a true or false whether the currentDate is valid or not. See selectable dates.
      */
-    isValidDate?: (x: string) => void;
+    isValidDate?: (x: any, y: any) => boolean;
```
simeg: Are `x` and `y` very good names for the parameters? In this case we know that `x` represents the `currentDate`, shouldn't we use that as the name of the parameter then instead? This should apply to all definitions in this file then.

react-datetime.d.ts
```diff
@@ -106,6 +106,18 @@ declare module ReactDatetime {
      When true, once the day has been selected, the react-datetime will be automatically closed.
      */
     closeOnSelect?: boolean;
+
```
simeg: Remove this empty line, please.

react-datetime.d.ts
```diff
@@ -106,6 +106,18 @@ declare module ReactDatetime {
...
+    /*
+    Allow to add some constraints to the time selector. It accepts an object with the format {hours:{ min: 9, max: 15, step:2}}
```
simeg: It seems like the styling of this block isn't correct, if you compare it with the other comment blocks. It should be pushed to the right a bit more. Also, please break the line manually so it doesn't have to break on it's own (look how long the lines are in other parts of this file). :)

react-datetime.d.ts
```diff
@@ -106,6 +106,18 @@ declare module ReactDatetime {
...
+    so the hours can't be lower than 9 or higher than 15, and it will change adding or subtracting 2 hours everytime the 
+    buttons are clicked. The constraints can be added to the hours, minutes, seconds and milliseconds.
+    */
+    timeConstraints: any; 
+    /*
+    When true, keep the picker open when click event is triggered outside of component. When false, close it.
+    */
+    disableOnClickOutside?:	boolean;
```
simeg: There's some extra space between `disableOnClickOutside?:` and `boolean;` that can be removed.

react-datetime.d.ts
```diff
@@ -106,6 +106,18 @@ declare module ReactDatetime {
...
     closeOnSelect?: boolean;
+
+    /*
+    Allow to add some constraints to the time selector. It accepts an object with the format {hours:{ min: 9, max: 15, step:2}}
+    so the hours can't be lower than 9 or higher than 15, and it will change adding or subtracting 2 hours everytime the 
+    buttons are clicked. The constraints can be added to the hours, minutes, seconds and milliseconds.
+    */
+    timeConstraints: any; 
+    /*
+    When true, keep the picker open when click event is triggered outside of component. When false, close it.
```
simeg: Same with this comment block, should be indented properly.

react-datetime.d.ts
```diff
@@ -106,6 +106,18 @@ declare module ReactDatetime {
...
+    */
+    disableOnClickOutside?:	boolean;
+
```
simeg: This line can be removed as well.

react-datetime.d.ts
```diff
@@ -106,6 +106,18 @@ declare module ReactDatetime {
      When true, once the day has been selected, the react-datetime will be automatically closed.
      */
     closeOnSelect?: boolean;
+
+    /*
+    Allow to add some constraints to the time selector. It accepts an object with the format {hours:{ min: 9, max: 15, step:2}}
+    so the hours can't be lower than 9 or higher than 15, and it will change adding or subtracting 2 hours everytime the 
+    buttons are clicked. The constraints can be added to the hours, minutes, seconds and milliseconds.
+    */
+    timeConstraints: any;
```
simeg: The `any` should be replaced with an `object`.

skyflyer:
Simon, I appreciate your comments. As I was just fixing them to have the right signature, I didn't think about actually improving the signatures of each definition, as `x` is as obvious to me as `date`, but your comments are spot on! Having said that, let me jump right in and address your comments

react-datetime.d.ts
skyflyer: I believe the documentation is phrased in a way that leads to confusion. If we look through source code, we see that the result of `onChange` is never used. The documentation string is taken from the root `README.md` file. I've amended that part of the configuration as well. Ref: `DateTime.js:184`.

skyflyer:
Simon:

* renderDay: I've changed the signature based on the doc snippet
```
    renderDay: function( props, currentDate, selectedDate ){
        return <td {...props}>{ '0' + currentDate.date() }</td>;
    },
```
* renderMonth: similar to renderDay above, but different signature
* renderYear: similar to above

README.md
```diff
@@ -42,9 +42,9 @@ render: function() {
...
 | **locale** | `string` | `null` | Manually set the locale for the react-datetime instance. Moment.js locale needs to be loaded to be used, see [i18n docs](#i18n).
-| **onChange** | `function` | empty `function` | Callback trigger when the date changes. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If it isn't, the value of the input (a `string`) is returned. |
-| **onFocus** | `function` | empty `function` | Callback trigger for when the user opens the datepicker. |
-| **onBlur** | `function` | empty `function` | Callback trigger for when the user clicks outside of the input, simulating a regular onBlur. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If it isn't, the value of the input (a `string`) is returned. |
+| **onChange** | function | empty function | Callback trigger when the date changes. The callback receives the selected `moment` object as only parameter, if the date in the input is valid. If the date in the input is not valid, the callback receives the value of the input (a string). |
```
simeg: Please stay consistent with the formatting of the types, `function` is a valid type so don't remove the backticks.

README.md
```diff
@@ -42,9 +42,9 @@ render: function() {
...
+| **onFocus** | function | empty function | Callback trigger for when the user opens the datepicker. |
```
simeg: (Same here) Please stay consistent with the formatting of the types, `function` is a valid type so don't remove the backticks.

README.md
```diff
@@ -57,6 +57,7 @@ render: function() {
 | **closeOnTab** | `boolean` | `true` | When `true` and the input is focused, pressing the `tab` key will close the datepicker.
 | **timeConstraints** | `object` | `null` | Add some constraints to the timepicker. It accepts an `object` with the format `{ hours: { min: 9, max: 15, step: 2 }}`, this example means the hours can't be lower than `9` and higher than `15`, and it will change adding or subtracting `2` hours everytime the buttons are clicked. The constraints can be added to the `hours`, `minutes`, `seconds` and `milliseconds`.
 | **disableOnClickOutside** | `boolean` | `false` | When `true`, keep the datepicker open when click event is triggered outside of component. When `false`, close it.
+>>>>>>> upstream/master
```
simeg: This should be removed.

simeg:
Hey @skyflyer. You're adding a lot of nice things, awesome! I just had some formatting comments, please look them over and I'll merge your changes :) (I still haven't figured out this new review thingy)

PR #149, opened 2016-09-16
Add name prop to enable normal HTML form use

mcorteel:
## Description
Just added a name prop.

## Motivation and Context
I would like to use this library in a normal HTML form.
I'm not sure how to build the component with webpack though (I use the dist version directly).

## Checklist
- [x] My change required changes to the documentation. (I guess)


simeg:
Hi, thanks for your contribution!

You should use this component with `npm` and `node`, so you install it (`npm install --save react-datetime`) and then require it in the file(s) you want to use the component in and `node` takes care of finding and importing it. We have examples in the README. You don't have to build it, we build it every time we release a new version so no need to worry about that.

In the checklist I see you have marked "_My change required changes to the documentation_". If your PR needs changes to the documentation, _you_ should be the one to update the documentation. The checklist is a reminder for people so they don't forget important stuff like that.

Finally, about the actual change. There's no need for a `name` prop because we have `inputProps`! You can input an object like `{ name: 'Arbitrary name' }` to the `inputProps` prop, which will attach the value (`'Arbitrary name'`) to the key `name` on the HTML form element (if that makes any sense) :)

mcorteel:
Oh, great! Sorry I didn't see that in the doc.
Thank you for the great work :)

PR #152, opened 2016-09-25
add clear date button

tunessofia:
## Description

**Added a button to the render of the DateTimePickerDays class. This button has a function "clearInput" that is triggered whenever we click in the button. This allows to clear the date input.
The function "clearInput" is passed by the parent class (DatePicker.js) as a props to the children, the parent holdes all the logic. This function just sets an empty input to a variable and sets the state of the "inputValue" with that variable.
This button is only shown when the "input" attrb is set to "true".
Also added a new dependecy to the package.json, the react-icons to style the button in the datepicker.**
## Motivation and Context

**This resolves a problem I was having in my project. When we add the "readOnly" attrb to the input element, clearing the date is not possible.
With this change is always possible to clear the date, except when it has the "disabled" attrb.**
## Checklist

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -192,6 +192,12 @@ var Datetime = React.createClass({
 		});
 	},
 
+	clearInput: function(){
+		var clearedDate = '';
+		this.setState({ inputValue: clearedDate });
+		return;
+	},
+
 	onInputKey: function( e ){
 		if ( e.which === 9 && this.props.closeOnTab ){
 			this.closeCalendar();
@@ -343,9 +349,9 @@ var Datetime = React.createClass({
 	},
 
 	componentProps: {
-		fromProps: ['value', 'isValidDate', 'renderDay', 'renderMonth', 'renderYear', 'timeConstraints'],
+		fromProps: ['value', 'isValidDate', 'renderDay', 'renderMonth', 'renderYear', 'timeConstraints', 'input'],
 		fromState: ['viewDate', 'selectedDate', 'updateOn'],
-		fromThis: ['setDate', 'setTime', 'showView', 'addTime', 'subtractTime', 'updateSelectedDate', 'localMoment']
+		fromThis: ['setDate', 'setTime', 'showView', 'addTime', 'subtractTime', 'updateSelectedDate', 'localMoment', 'clearInput']
 	},
 
 	getComponentProps: function(){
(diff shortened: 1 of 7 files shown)


simeg:
Thanks for your contribution @tunessofia.

Why do you need the `react-icons` dependency? This module should try to avoid dependencies, especially those that do not affect a lot of things. I do not think `react-icons` qualifies as one of those dependencies.

You say you cannot use `readOnly` and at the same time clear the input. Have you tried using the `value` prop and setting it to just empty string? 

(Also, please use the checklist as well, it's there for a reason :) )

tunessofia:
@simeg That's the thing, when it's readOnly I dont have any button on the datepicker so the user can clear the input if needed. 

The only reason for the react-icons dependency was just a way to avoid the text in the button and to avoid the translation part. But can totally be substituted by something else.

(will do :))

simeg:
@tunessofia but can't you have a button outside of this component that sets the value of this component to empty string (if that makes any sense..)?

tunessofia:
@simeg totally, but imo I think it makes more sense if the datepicker has that functionality built in, than obligate the developer to implement such feature. It's like the user experience when we select a date in the calendar, with the right property set the calendar will close when we click. But this might be something that can be added or resolved with the last issue you opened. Having a button inside the input element, at the left or right margin.
Although would be important in that case to have the possibility of an input that doesnt allow the user to manually write but allow it to interact with the button.
The main problem for me is having an input that allows the user to write, and thats the reason I put it readOnly, but then if the user wants to clear the date, he cant. But in spite of that, a clear button is essencial imo.
This is just a suggestion for an approach of course, i will probably use this as my fix for now but i think its a nice feature for you to consider for the future.

simeg:
@tunessofia I hear you, I never really thought about the impact of applying `readOnly`. 

To me it is important that this component keeps it's generic touch, and that's why I think #153 is a good idea. About your problem, I did some quick testing when applying `readOnly` and I _think_ it can be done by using the method I described earlier - having a button set the `value` to empty string. If I find the time I will investigate further and report back.

Thanks again for your contribution. I wont merge this right now but I'll leave it open for a while for discussion.

tunessofia:
Yes, I totally agree with the #153 ideia and makes total sense what the @arqex wrote in the #134. Thanks @simeg !

simeg:
Closing this, let's continue any discussion in #153.

Issue #153, opened 2016-09-26
Enable pre and post content to be added


simeg:
Reading the comments for #45 there's a suggestion to add prefix and postfix properties to enable to input content on either left (pre) or right (post) side of the input field. 

To me this seems like a reasonable approach to enable users to add e.g. icons, buttons and similar things (that's often requested). This way we're not changing any code of the actual component.

A downside to this is of course that the CSS could break due to changing the HTML, but that is a risk I think is OK.

I'm interested in hearing people's opinions about this (especially @arqex's).

arqex:
Hey @simeg 

I had thought something even more generic. We can have a `renderInput` prop that accepts a function like:

```js
function renderInput( inputProps ){
 return <input { ...inputProps } />
}
```

That would print the input just right like it is now, so it would be our default `renderInput` prop. But that is so flexible, people could customize it so much that they could render whatever they like, not even an input there.

We could do the same with the views, `renderView` would receive the type of the view to render:

```js
function renderView( viewType, ViewComponent, viewProps ){
 return <ViewComponent { ...viewProps } />
}
```

Again the developer could completely customize every view (years, months, days, time), changing it slightly or building a completely new view.

simeg:
This was implemented in #464, looks like we didn't have to wait for v3!

PR #158, opened 2016-10-04
Add handling custom input component

rroslaniec:
## Description
It's allow to change default input component with complex ones.

## Motivation and Context
More complex customization on input.

## Checklist

- - [x] I have updated the documentation accordingly.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -28,6 +28,7 @@ var Datetime = React.createClass({
 		onChange: TYPES.func,
 		locale: TYPES.string,
 		input: TYPES.bool,
+		inputComponent: TYPES.element,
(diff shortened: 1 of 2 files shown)


simeg:
Hi @rroslaniec, thank you for your contribution. I like this PR (besides the fact that there's no tests included), you even wrote an example - nice! I agree with you that you should be able to use your own custom input element with this component, but if you look at #153 you'll see how we will implement that functionality. To me it makes sense to follow the same pattern as we currently have with `renderDay`, `renderMonth` and `renderYear`. 

I will close this as I won't merge it, I hope you agree with me. Thank you.

PR #159, opened 2016-10-05
Signpost type definition file in package.json

leth:
Without this option typescript doesn't find the current type definitions automatically.

For more info see the [docs for publishing type definitions](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).


simeg:
Nice catch, thanks @leth!

PR #160, opened 2016-10-11
Added boolean property blockInputEdit

arturd-softgrad:
## Checklist

- [x] I have updated the documentation accordingly.
- [x] I have updated the TypeScript type definition accordingly.

If set to true, then prohibits the input of dates in the input field with the keyboard


arturd-softgrad:
Could you update the npm package and tell me about it ?

simeg:
Hi @arturd-softgrad, thanks for your contribution. I like this idea of not allowing the user to edit the input field, and I think it should be included in the component. We do however have plans on opening up for the developer to customise the input element, and then the developer could set the `disabled` attribute to create this effect. That makes more sense to me - to do it the "native" way. 

I will close this as I won't merge it. As I said I really like this idea, but I agree with @arqex in issue #153 and as we have the the props `renderDay`, `renderMonth` and `renderYear` it makes sense in also editing the input element the same way. Thanks.

PR #161, opened 2016-10-12
Disable months and years according to isValidDate

### LINKED ISSUE #129: IsValidDate Not hire when dateFormat is set
Hi,
The function IsValidDate is not hire when we set a dateFormat
example : http://codepen.io/anon/pen/dpyEGr?editors=1112

In the example, we want limit month selector, without this function we can't...

michelre:
Disable dates according to isValidDate function when the viewMode and dateFormat props are set.
## Description

My solution was to add a disabled class when the isValidDate is not satisfied and thus allowing the click accordingly!
## Motivation and Context

This should be fixed #129 
## Checklist

### DIFF
--- a/example/example.js
+++ b/example/example.js
@@ -3,6 +3,12 @@ var React = require('react');
 var ReactDOM = require('react-dom');
 
 ReactDOM.render(
-  React.createElement(DateTime, { timeFormat: true }),
+  React.createElement(DateTime, {
+    viewMode: 'months',
+    dateFormat: 'MMMM',
+    isValidDate: function(current){
+      return current.isBefore(DateTime.moment().startOf('month'));
+    }
+  }),
   document.getElementById('datetime')
 );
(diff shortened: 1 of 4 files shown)


simeg:
Hi @michelre, thanks for your contribution! It would be awesome if this PR would solve the bug. I'm gonna check this out when I have more time, maybe tomorrow. If you have time and energy to write one or more more tests the process would move a long a bit quicker :)

simeg:
@michelre Sorry for the delay, I'll try to find some time to look at this soon.

simeg:
Thanks for this great contribution. I like this PR because it has many small commits - exactly how it should be done. It makes it so much easier for me to review.

This is a requested feature so I think a lot of people will be happy about this change 🎉.

michelre:
That's cool, I really hope this feature's gonna help people. I had so much pain to find a great react date picker library. Thanks your work! :)

PR #165, opened 2016-10-12
fix #163 componentWillReceiveProps() is not honouring old state

### LINKED ISSUE #163: componentWillReceiveProps() is not honouring old state
when `currentView = 'days'` and `update.open === undefined` , if the datepicker state was `open`, it should remain `open`. In other words, when `currentView = 'days` and the parent component re render executing datepicker `componentWillReceiveProps()`, datepicker will close.

```
if ( update.open === undefined ){
            if ( this.props.closeOnSelect && this.state.currentView !== 'time' ){
                update.open = false;
            }
            else {
                update.open = this.state.open;
            }
        }
```

gvelo:
fix #163 componentWillReceiveProps() is not honouring old state


just-boris:
@simeg ping.

I have also encountered with the same problem, this PR seems like it resolves that.

simeg:
@gvelo Hey, sorry about the delay for the response for this. As this PR changes an important piece of logic and does not break any test I didn't feel confident in merging it, now I've done some investigation and it turns out this change is a very good one! Thanks a lot.

Issue #166, opened 2016-10-13
Displayed selected date not updated when dateFormat prop changes


MilosRasic:
See title.

My guess is the problem is in `componentWillReceiveProps()`. A quick dirty edit in node modules confirms that doing a full `this.getStateFromProps(nextState)` in `DateTime.js` line 161 solves the problem.

PR #167, opened 2016-10-13
Fixed the problem with displayed selected datetime not changing when the format props change

MilosRasic:
## Description

In case of dateFormat and/or timeFormat prop change, componentWillReceiveProps was updating only the inputFormat state, not updating the inputValue and leaving it in old format. The value would correctly updated to the new format when the user manually changes the selected datetime.
## Motivation and Context

Fixes https://github.com/YouCanBookMe/react-datetime/issues/166
## Checklist

Added example/formatChangeExample.js as demo for the format change. To run it simply replace example.js with formatChangeExample.js in example/webpack.config.js. Date format should change after 3 seconds.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -154,11 +154,8 @@ var Datetime = React.createClass({
 			update = {}
 		;
 
-		if ( nextProps.value !== this.props.value ){
+		if ( nextProps.value !== this.props.value || formats.datetime !== this.getFormats( this.props ).datetime ){
 			update = this.getStateFromProps( nextProps );
-		}
-		if ( formats.datetime !== this.getFormats( this.props ).datetime ) {
-			update.inputFormat = formats.datetime;
 		}
 
 		if ( update.open === undefined ){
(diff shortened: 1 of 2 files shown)


simeg:
Hi @MilosRasic and thanks for your contribution. Sorry for the late response.

I'll review this as soon as possible.

simeg:
I tried writing a test for this, but I came to the conclusion that our current testing methods doesn't support changing props on the go. So a bigger effort would require to write this test, maybe you came to the same conclusion and decided to "test" it using an example? 

Anyway, the PR looks good to me. Thanks for your contribution!

MilosRasic:
> "test" it

Wow, you wound me :)

The example is intended to demonstrate the issue since you can't do it with regular user interaction with the component using the existing example.

I'm still quite unsure about testing in React, to be honest. This seemed to be something I wouldn't want to test but now that I think about it it could be broken by future changes. I'll look into submitting a PR for tests as soon as I have some time.

simeg:
Sorry, it's a proper and completely valid test!

I found an article about testing in React which you can find in the [Backlog I just created](https://github.com/YouCanBookMe/react-datetime/projects/2) for this project. I scanned it quickly and it looked like something we could benefit from, to actually change the state from outside the component.

It would be awesome it you could find some time for this. Looking forward to future PRs, thanks!

PR #172, opened 2016-10-20
True stateless mode when prop `open` specified

vshushkov:
## Description
True stateless mode when prop `open` specified.

## Motivation and Context
Internal component's state is resetting when a component is updating. The previous behavior of property `open` not working properly. Because of this picker is suddenly closing when it wasn't necessary to do (ex. after updating a redux state).
Now, when prop `open` is provided the stateless mode is turning on. So, when `Datetime` think if a widget has to be shown or closed it calls methods onOpen/onClose to let you decide how exactly prop `open` need to be changed.

## Checklist

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -26,6 +26,8 @@ var Datetime = React.createClass({
 		onFocus: TYPES.func,
 		onBlur: TYPES.func,
 		onChange: TYPES.func,
+		onOpen: TYPES.func,
+		onClose: TYPES.func,
 		locale: TYPES.string,
 		input: TYPES.bool,
 		// dateFormat: TYPES.string | TYPES.bool,
@@ -50,6 +52,8 @@ var Datetime = React.createClass({
 			onFocus: nof,
 			onBlur: nof,
 			onChange: nof,
+      onOpen: nof,
+      onClose: nof,
 			timeFormat: true,
 			timeConstraints: {},
 			dateFormat: true,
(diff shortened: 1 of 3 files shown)


vshushkov:
Example with `redux`:

```js
import React from 'react';
import { connect } from 'react-redux';
import { close, open } from '../../actions/forms/datetimepicker';
import Datetime from 'react-datetime';

const DatetimePicker = ({ opened, openCalendar, closeCalendar, ... }) => (
  <Datetime
    open={opened}
    onOpen={openCalendar}
    onClose={closeCalendar}
    ...
  />
);

const mapStateToProps = (state, { field }) => ({
  opened: (state.datetimepickers[field.name] || {}).opened === true
});

const mapDispatchToProps = (dispatch, { field }) => {
  return {
    closeCalendar: (event) => dispatch(close(field.name)),
    openCalendar: () => dispatch(open(field.name)),
    ...
  }
};

export default connect(
  mapStateToProps,
  mapDispatchToProps
)(DatetimePicker);
```

simeg:
Hi @vshushkov, thanks for your contribution! Sorry for the late response.

I've looked at the code and the tests and I'm pleased with the quality. I like everything about this PR actually, but what concerns me is possibly breaking the API. I agree with you that it's broken in that sense that it's not truly describing the state of the component.

Let's say we merge this, then people will be affected because we change how the `open` works. Me and the other maintainer have initiated discussions about version 3.0 of this component. I think this would be a great change for that major version bump. The `.editorconfig` file I think I'll just add outside of this PR (after testing it). Or if you'd like to create a new PR, that is up to you =)

Do you agree that it will break the API because we're changing the way `open` works?

vshushkov:
Hi @simeg! Thanks. Yes, I think it's a right decision. I'll remove `.editorconfig` from PR.

PR #180, opened 2016-11-02
added new props isValidMonth to enable or disable the particular month

sridharAJ:
Added new props to validate the month view

## Description
We have month only view. For this option I want to disable or enable click on months based on some condition. For this I added isValidMonth props. (Similar like isValidDate).

## Motivation and Context
This should be merged, because its mandatory if any want to disable and enable in month view based on condition.
Yes it does, Now we have control to enable and disable even in Month only view.

## Checklist

- [ x] My change required changes to the documentation.
- [x ] All new and existing tests passed.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -7,7 +7,7 @@ var assign = require('object-assign'),
 	YearsView = require('./src/YearsView'),
 	TimeView = require('./src/TimeView'),
 	moment = require('moment')
-;
+	;
 
 var TYPES = React.PropTypes;
 var Datetime = React.createClass({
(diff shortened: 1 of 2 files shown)


simeg:
Hi @sridharAJ, thanks for your contribution! Sorry for the late response.

I like this idea. But your commit contains a lot of white space changes, fix that please. I'd also ask you to write tests for your change, but I see that we don't have any tests for the current `isValidDate` :/. Can you write at least one test for your changes? I'd really feel more confident in merging it, and because this change isn't breaking the API in any way we could merge it and release a new version.

simeg:
I decided to merge a PR using the same prop (`isValidDate`) to control everything (date, month, year), so the functionality this PR is adding has already been added in https://github.com/YouCanBookMe/react-datetime/commit/c306f2fcc9b08aef38a246f94949826965ff66c1. 

To me it makes more sense to have one prop for this than to have separate props for each "unit". Thanks for your contribution though! :relaxed:

PR #182, opened 2016-11-09
Add support for inputting times as UTC

deleted:
## Description
This change adds a boolean `utc` prop. When set to `true`, it will cause moment to interpret input times as UTC, rather than the user's local timezone.

## Motivation and Context

This PR will help anyone who deals with timestamps primarily in UTC, for example anyone operating in UTC or across multiple timezones.

I build tools to manage a fleet of satellites in low earth orbit, where local time zones don't make any sense. All of our data is stored and queried as UTC. This change will allow my users to query the dataset using its native timezone.

## Checklist

- [x] I have updated the documentation accordingly.
- [x] I have updated the TypeScript type definition accordingly.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -27,6 +27,7 @@ var Datetime = React.createClass({
 		onBlur: TYPES.func,
 		onChange: TYPES.func,
 		locale: TYPES.string,
+		utc: TYPES.bool,
 		input: TYPES.bool,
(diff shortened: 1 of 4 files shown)


simeg:
Hi @deleted. Thanks for your contribution!

I wish all PRs looked like this, small, compact and straight to the point. It makes my life so much easier :)

Issue #171, opened 2016-10-19
[Request] Add a direction prop to set which way the picker opens


MrLeebo:
This is a new feature request. If you position the Datetime control towards the bottom of the page, the picker will appear below it and off the page entirely, forcing the user to scroll before they can make a selection. I hope it would not be too much trouble to add a `direction` prop which could take `oneOf(['up', 'down', 'left', 'right'])` and position the picker in that direction relative to the input.

```
<Datetime direction="up" />
```

simeg:
Hi @MrLeebo. I've also had the specific problem you are mentioning, you can solve it using JS. Maybe even HTML using an anchor? But about your suggestion, I think the idea is good. If you'd like to create the code I'd be happy to review the PR.

MrLeebo:
I created the issue because I don't think I have time to learn the repo and implement this right now, but if that changes I will open up a PR.

danosborne:
+1

Will also try to spin up a PR...

simeg:
A tip: use react-date-playground for development (check the README).

danosborne:
Hacked for 5 mins and opened the PR above. Are people happy with the approach? 

If so I will tidy it up and actually follow the contributing guidelines etc. ;-)

Krzyku:
For me `direction` prop is misleading. I understand direction as left-to-right or right-to-left settings.
Maybe `position` is better idea for name of this feature?

bjbrewster:
I agree. Bootstrap Popover/Tooltip/Overlay uses prop `placement` with values `top`, `bottom`, `left` and `right`. Can we use the same prop name with `top` and `bottom`?

simeg:
`placement` sounds good to me. Will try to find time for looking into this soon.

kostiantyn-solianyk:
any updates?

Bramgus12:
Is there a workaround to get the date picker above the input field?

jvnlwn:
To move the date picker above the input, just add `bottom: 100%;` style to the `. rdtPicker` class. (Caveat in case it matters: this is version 2.14.0 pictured.)

<img width="269" alt="Screen Shot 2022-09-06 at 11 19 39 AM" src="https://user-images.githubusercontent.com/5227102/188673179-af6092c7-6eff-450e-ab37-018d83ef920f.png">

PR #184, opened 2016-11-09
Open the date picker upwards using direction prop

danosborne:
## Description
An inital 5-minute attempt at providing the enhancement requested in issue 171.
Just an initial request for feedback. Is the approach acceptable or awful?

## Motivation and Context
- https://github.com/YouCanBookMe/react-datetime/issues/171

Motivation for the PR is to get the ball rolling. TBH, this enhancement is enough for me.

However, I still need to add to the doco and write a test.

I can also see alternative approaches, all more laborious:

1. Ideally there'd be a direction: 'youKnowBest' option in which the component looks around its bounds to decide which way is best to open.
2. I think there's viable CSS-only approach to the 'up' case. 

What's your preference?

## Checklist
- [Y] My change required changes to the documentation.
- - [N] I have updated the documentation accordingly.
- - [N] I have updated the TypeScript type definition accordingly.
- [N] I have added tests to cover my changes.
- [Y] All new and existing tests passed.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -2,6 +2,7 @@
 
 var assign = require('object-assign'),
 	React = require('react'),
+  ReactDOM = require('react-dom'),
 	DaysView = require('./src/DaysView'),
 	MonthsView = require('./src/MonthsView'),
 	YearsView = require('./src/YearsView'),
@@ -37,7 +38,8 @@ var Datetime = React.createClass({
 		open: TYPES.bool,
 		strictParsing: TYPES.bool,
 		closeOnSelect: TYPES.bool,
-		closeOnTab: TYPES.bool
+		closeOnTab: TYPES.bool,
+    direction: TYPES.oneOf(['down', 'up'])
 	},
 
 	getDefaultProps: function() {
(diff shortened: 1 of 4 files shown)


simeg:
Hi @danosborne and thanks for your contribution.

If I were to implement this I'd probably do something similar as this (not saying it's the best way though). We are soon gonna start working on a V3 of this component and I think this would be a good feature for V3. I'll label this as a "candidate for V3" and I may or may not come back to it. I'll leave it open for now :) Thanks again for your contribution.

modularcoder:
+1 for this feature! :)

aridlehoover:
Why not just use CSS?

manuelmager:
+1

eug48:
I'm not sure why a CSS-based approach is more laborious?

```css
.rdtPickerOpenUpwards .rdtPicker {
  top: auto;
  bottom: 100%;
}
```

```js
<Datetime className="rdtPickerOpenUpwards" />

treetopvt:
eug48's CSS suggestion is good, but the arrow indicator in the popup no longer points to the source input (remains on the top). +1 on the position prop being added.

Issue #181, opened 2016-11-04
Typings not compatible with Typescript 2.0


AaronAsAChimp:
The typings for React have changed with Typescript 2.0. They are now installed through the NPM package '@types/react'. This causes a compilation failure, shown below, because there is no longer a  `__React` namespace.

New typings:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/react/index.d.ts

Error:
```
TypeScript error: [...]/node_modules/react-datetime/react-datetime.d.ts(8,18): Error TS2503: Cannot find namespace '__React'.
```

leth:
It might be enough just to change `__React` to `React`, can anyone confirm?

AaronAsAChimp:
I think you can just get rid of the whole import statement.

simeg:
I don't work with TypeScript so if someone with more experience than me (and an environment to actually test it in) could take a look at this I'd would be grateful. Thanks.

AaronAsAChimp:
I can submit a pull request with the typings I'm using, If you would find that helpful.

leth:
Perhaps we could also add a typescript example and make it part of the travis build? That would probably help @simeg in the long run.

grushetsky:
@AaronAsAChimp, could you share or submit your typings, please?

PeterKottas:
Hi guys, nice work on this issue, I gave it a go and it fixes the problem for us as well. I was just wondering if you know when this is going out. I have a guy waiting for PR because of the typings bug and not sure if I should just merge it in and create an override for the module somehow or wait if the release is only couple days away. Any info much appreciated, thanks :)

PeterKottas:
Ok fixing it myself in the meantime :) Please let me know when you release this one. Thanks ;)

ncphillips:
I am also awaiting this release!

PeterKottas:
Hi @ncphillips I suggest looking at #215, vissermc's answer, going to implement it myself at the moment as a temp fix.

simeg:
Closing this as #190 was merged and version 2.8.0 was just released.

PeterKottas:
Hi @simeg this still appears to be a problem in the recent release. Seems like you might have forgotten to merge. Typings are still wrong. Take a look at https://github.com/vissermc/react-datetime.git#512bf6f7c3f664b8268de9b3d68e4671feda903c which is working for me. You just need to comment out the line with react basically.

[...]

PR #190, opened 2016-11-17
Update typings for Typescript 2.0

AaronAsAChimp:
## Description
Typescript 2.0 has a new mechanism for resolving types for modules. This PR updates the typings for react-datetime to be compatible with with the latest typescript. It also includes a couple of fixes and enhancements to the typings.

- Restrict `viewMode` to the correct strings (or a number).
- Improve typings for `timeConstraints`
- Fix `renderDay`, `renderMonth`, `renderYear` to the correct return types.
- Improve typings for `inputProps`, limiting it to only those allowed on input elements
- Fix `ReactDatetime` to be a class

## Motivation and Context
Fixes issue #181. Improves compatibility with this package.

## Checklist
- [X] My change required changes to the documentation.
- - [X] I have updated the documentation accordingly.
- - [X] I have updated the TypeScript type definition accordingly. :P
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,14 +1,35 @@
 // Type definitions for react-datetime
 // Project: https://github.com/YouCanBookMe/react-datetime
 // Definitions by: Ivan Verevkin <vereva@x-root.org>
+//     Updates by: Aaron Spaulding <aaron@sachimp.com>
 
-//// <reference path="../moment/moment-node.d.ts" />
+declare module "react-datetime" {
 
-declare module ReactDatetime {
-  import React = __React;
-  // import * as moment from 'moment';
+  /*
+   A stand-in type for Moment, this file currently has no way of guaranteeing
+   the existence of those typings.
+   */
+  type Moment = any;
 
-  export interface DatetimepickerProps {
+  /*
+   The view mode can be any of the following strings.
+   */
+  type ViewMode = 'years' | 'months' | 'days' | 'time';
+
+  interface TimeConstraint {
+    min: number;
+    max: number;
+    step: number;
+  }
+
+  interface TimeConstraints {
+    hours?: TimeConstraint;
+    minutes?: TimeConstraint;
+    seconds?: TimeConstraint;
+    milliseconds?: TimeConstraint;
+  }
+
+  interface DatetimepickerProps {
     /*
      Represents the selected date by the component, in order to use it as a controlled component.
      This prop is parsed by moment.js, so it is possible to use a date string or a moment.js date.
@@ -46,15 +67,11 @@ declare module ReactDatetime {
      */
     locale?: string;
     /*
-     Whether to interpret input times as UTC or the user's local timezone.
-     */
-    utc?: boolean;
-    /*
      Callback trigger when the date changes. The callback receives the selected `moment` object as
      only parameter, if the date in the input is valid. If the date in the input is not valid, the
      callback receives the value of the input (a string).
      */
-    onChange?: (momentOrInputString: string|any) => void;
+    onChange?: (momentOrInputString: string|Moment) => void;
     /*
      Callback trigger for when the user opens the datepicker.
      */
@@ -65,11 +82,11 @@ declare module ReactDatetime {
      is valid. If the date in the input is not valid, the callback receives the value of the
      input (a string).
      */
-    onBlur?: (momentOrInputString : string|any) => void;
+    onBlur?: (momentOrInputString : string|Moment) => void;
     /*
      The default view to display when the picker is shown. ('years', 'months', 'days', 'time')
      */
-    viewMode?: string|number;
+    viewMode?: ViewMode|number;
     /*
      Extra class names for the component markup.
      */
@@ -77,7 +94,7 @@ declare module ReactDatetime {
     /*
      Defines additional attributes for the input element of the component.
      */
-    inputProps?: Object;
+    inputProps?: React.HTMLProps<HTMLInputElement>;
     /*
      Define the dates that can be selected. The function receives (currentDate, selectedDate)
      and should return a true or false whether the currentDate is valid or not. See selectable dates.
@@ -88,21 +105,21 @@ declare module ReactDatetime {
      the selectedDate, the current date and the default calculated props for the cell,
      and must return a React component. See appearance customization
      */
-    renderDay?: (props: any, currentDate: any, selectedDate: any) => React.Component<any, any>;
+    renderDay?: (props: any, currentDate: any, selectedDate: any) => JSX.Element;
     /*
      Customize the way that the months are shown in the month picker.
      The accepted function has the selectedDate, the current date and the default calculated
      props for the cell, the month and the year to be shown, and must return a
      React component. See appearance customization
      */
-    renderMonth?: (props: any, month: number, year: number, selectedDate: any) => React.Component<any, any>;
+    renderMonth?: (props: any, month: number, year: number, selectedDate: any) => JSX.Element;
     /*
      Customize the way that the years are shown in the year picker.
      The accepted function has the selectedDate, the current date and the default calculated
      props for the cell, the year to be shown, and must return a React component.
      See appearance customization
      */
-    renderYear?: (props: any, year: number, selectedDate: any) => React.Component<any, any>;
+    renderYear?: (props: any, year: number, selectedDate: any) => JSX.Element;
     /*
      Whether to use moment's strict parsing when parsing input.
      */
@@ -117,7 +134,7 @@ declare module ReactDatetime {
      it will change adding or subtracting 2 hours everytime the buttons are clicked. The constraints
      can be added to the hours, minutes, seconds and milliseconds.
     */
-    timeConstraints?: Object;
+    timeConstraints?: TimeConstraints;
     /*
      When true, keep the picker open when click event is triggered outside of component. When false,
      close it.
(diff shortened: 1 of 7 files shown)


typings/index.d.ts
```diff
@@ -46,15 +67,11 @@ declare module ReactDatetime {
      */
     locale?: string;
     /*
-     Whether to interpret input times as UTC or the user's local timezone.
-     */
-    utc?: boolean;
```
simeg: Why was this prop removed?

typings/index.d.ts
```diff
@@ -1,14 +1,35 @@
 // Type definitions for react-datetime
 // Project: https://github.com/YouCanBookMe/react-datetime
 // Definitions by: Ivan Verevkin <vereva@x-root.org>
+//     Updates by: Aaron Spaulding <aaron@sachimp.com>
 
-//// <reference path="../moment/moment-node.d.ts" />
+declare module "react-datetime" {
```
simeg: Here you are using double-quotes but on L17 you are using single-quotes. Does the module declaration require double-quotes? If not I think you should change them to stay consistent :)

typings/index.d.ts
```diff
@@ -1,14 +1,35 @@
...
+
+  /*
+   A stand-in type for Moment, this file currently has no way of guaranteeing
+   the existence of those typings.
```
simeg: I like comments like this :relieved:

typings/react-datetime-tests.tsx
```diff
@@ -0,0 +1,170 @@
+import * as React from 'react';
+import { Moment } from 'moment';
```
simeg: This file is written with ES6 syntax. I prefer ES6 syntax, but none of the other files are written in this syntax. My first thought is that we should stay consistent, but at the same time this could be the first step to making everything into ES6 (which I plan on doing eventually). What are your thoughts on this?

typings/react-datetime-tests.tsx
```diff
@@ -0,0 +1,170 @@
...
+ */
+
+const TEST_CUSTOMIZABLE_COMPONENT_PROPS: JSX.Element = <ReactDatetime
+		renderDay={ (props, currentDate, selectedDate) => {
+			return <td {...props}>{ '0' + currentDate.date() }</td>;
+		} }
+		renderMonth={ (props, month, year, selectedDate) => {
+			return <td {...props}>{ month }</td>;
+		} }
+		renderYear={ ( props, year, selectedDate ) => {
```
simeg: You have spaces in the arguments here that you don't have on L118 or L121 (sorry if this is a bit picky).

typings/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+        "typeRoots": [
+            "../"
+        ],
+        "types": [],
+        "noEmit": true,
+        "jsx": "react"
+    },
+    "files": [
+        "index.d.ts",
+        "react-datetime-tests.tsx"
```
simeg: If moving the test file this reference will have to be updated as well (just a friendly reminder).

simeg:
Hey @AaronAsAChimp, thanks for this awesome contribution! As you can see I added some comments to the code.

These are breaking changes for people using version < 2.0 right? Is there some way to make it compatible for both, or has the whole community already embraced version 2?

typings/index.d.ts
AaronAsAChimp: ... a mistake.

typings/react-datetime-tests.tsx
```diff
@@ -0,0 +1,170 @@
+import * as React from 'react';
```
AaronAsAChimp: I was following the [recommendations](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package) from the DefinitelyTyped repository. This one doesn't necessarily apply here, I can move it.

typings/react-datetime-tests.tsx
AaronAsAChimp: I prefer using ES6 syntax, the code is cleaner and allows for better static analysis. It is also the preferred style for Typescript.

AaronAsAChimp:
You're correct they are breaking changes. DefinitelyTyped maintains a separate branch for types-2.0 to preserve the 1.8 typings. I wasn't able to find a way to make these typings work for both.

typings/react-datetime-tests.tsx
simeg: All right, let's go with ES6.

typings/react-datetime-tests.tsx
simeg: So in the DefinitelyTyped readme it says

> **File**: foo-tests.ts
> **Purpose**: This contains sample code which tests the typings. This code does not run, but it is type-checked.

And that's what's happening in your changes, right? I didn't get that at first but yeah if strictly the typings are being tested I can see them living in the folder they're in now. 

I see the typings being a separate modular thing on top of the component, so if this is the case (please confirm) then I'd say just leave it as is.

AaronAsAChimp: Yes, that is definitely the case 😄

simeg:
I am trying to figure out if we can support both v1.8 and v2.0, but I can't really see a way where we don't have to maintain two branches/repos, something I do not want to do.

I am leaning towards merging this, because I like the tests and when looking at the breaking changes from 1.8 -> 2.0 there's not really a lot of them. @AaronAsAChimp what are your thoughts? Any idea on how we can support both versions?

vutran:
+1 for the TS 2.0 support. If people are still on TS 1.8, they most like are not ahead in updating their packages as well (in this case, `react-datetime`).

simeg:
@vutran that is true, but usually dependencies are marked with a carrot (`^`) which means it wont update to the next major version but minor and patch versions are ok. This PR would be a minor version bump for `react-datetime` but the TS version bump 1.8 => 2.0 is a major.

Maybe I'm overthinking this, I just don't want to answer to people asking why we broke their application. I welcome every opinon on this matter :)

schickling:
Would be great if @vutran could provide this via `@types` until this is merged. As in fact it's breaking our build since we're using Typescript 2.1.

AaronAsAChimp:
I would prefer we either:

1. Commit to having the typings here for the latest typescript and move the 1.8 typings and any future legacy typings to DefinitelyTyped, or
2. move all typings there and leave this repository to be JavaScript only.

Having them in both places would only lead to confusion, extra maintenance, and erroneous bug reports when someone inadvertently uses stale typings. Even if its only temporary. But Its not really up to me, its up to the maintainers and how much time they are willing to spend supporting these typings.

simeg:
@AaronAsAChimp I like the first option you are suggesting. In a few months we could look into only using v2.0+ in both DefinitelyTyped and here.

@AaronAsAChimp @vutran @schickling I am reading the instructions in the README for the DefinitelyTyped repo, and I'm getting the feeling that some of you people that actually _use_ TS can publish the 1.8 version quicker and better. Does anyone feel up for it? 

In what order do we do this? Can we merge this now and publish the 1.8 version to DefinitelyTyped later?

Thanks so much for the help and input everyone!

typings/react-datetime-tests.tsx
```diff
@@ -0,0 +1,171 @@
...
+import { Moment } from 'moment';
+import ReactDatetime = require('react-datetime');
```
bovan: this should be equal to
`import * as ReactDatetime from 'react-datetime';`

but it's not.. i get the following error:
`[ts] Module ''react-datetime'' resolves to a non-module entity and cannot be imported using this construct.`

I now get no errors in vscode using `import ReactDatetime from 'react-datetime';` , but its not able to import it properly so it breaks during build

is it possible to make it work using the `import * as name` syntax?

simeg: @AaronAsAChimp would you be able to answer this? Thanks so much.

AaronAsAChimp: Sorry it took so long to reply, but I wanted to make sure I had it right. They are not equivalent.

The following is for CommonJS modules that use `module.exports = Thing`. This may be an ugly way of importing, but it is not within the scope of this PR to change it. It would also massively break the way non-typescript users require the module.

```ts
import ReactDatetime = require('react-datetime');
```

When you have multiple named exports `module.exports = { Thing: Thing }` or `export  class Thing { ... }` it is imported like so.

```ts
import * as ReactDatetime from 'react-datetime';
```

Finally, when you have a default export: `export default class Thing {...}`

```ts
import ReactDatetime from 'react-datetime';
```

You can read more here
https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

simeg:
About this PR - I want to wait to see what happens to [the PR in DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13022). If it gets merged I will merge this but if not I will have to think about the next step.

typings/react-datetime-tests.tsx
bovan: @AaronAsAChimp @simeg I suspect the documentation isn't up to date.. the require syntax is what typescript implemented before the ES6 syntax was ready.

The main issues here are: 
- Editors does not allow you to type `import * as ReactDatetime from 'react-datetime';` which works
- Editors allows you now to type: `import ReactDatetime from 'react-datetime';` which doesnt work.

This indicates that there is something wrong with the typings. (Also all other imports we use have no problems with the import * as- syntax, and there's a lot).
I'm sure there is a way to make the typings work without rewriting the actual code itself, unfortunately I've not had time to do further research into this.

Anyway, the problem for us is that we use tslint with `"no-var-requires": true`, but i work around that with disable-line, so it's not game breaking for me, just a heads up..

AaronAsAChimp:
With the DefinitelyTyped pull request closed. Is this ready for merge?

I've updated the branch with the suggested changes.

simeg:
Sorry for the late response everyone. So we did end up with supporting both 1.8 and 2.0+ in the same repo. I am ok with that if that's the suggestion from the DefinitelyTyped people, so I'll go ahead and merge this PR.

Thanks a lot everyone who contributed and a **big thank you** to @AaronAsAChimp for everything.

Issue #192, opened 2016-11-17
isValidDate is hiding current month and year


simeg:
1) [Look at this demo](http://codepen.io/simeg/pen/XNNYJg)
2) Notice the current month (November right now)
3) Notice the current year (2016 right now)
4) Click "November 2016" to go to month view
5) November is a disabled option **<- BUG**
6) Click "2016" to go to year view
7) 2016 is a disabled option **<- BUG**

As the functionality of actually validating months and year was introduced pretty recently, I'm guessing it was [this commit](c306f2fcc9b08aef38a246f94949826965ff66c1) that caused it.

@michelre do you feel like looking into it? :)

michelre:
Thanks for your feedback. I'll check this out this week end! :)

PR #198, opened 2016-11-19
Disable months years fix 192

michelre:
isValidDate bug fix

## Description
The isValidDate pull request wasn't fully functional and caused some bugs

## Motivation and Context
Should fix the bug #192 

## Checklist
- [ ] I have updated the documentation accordingly.
- [ ] I have updated the TypeScript type definition accordingly.


simeg:
Hi @michelre. Thanks for looking into this bug!

Before I review it you need to rebase your branch against the master branch. [Here's an SO post](http://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-with-remote-master) on how you do it, in case you didn't know :)

scags9876:
I'm running into this issue and have had to pin react-datetime to version 2.6.2 until this is fixed.

I think the fix in this PR is pretty great, but there is still a small hole in the logic.  Lets say my isValidDate function checks if the date is after noon on the last day of the month.  The month will still be disabled in that case.

Would it be a good idea to pass the viewMode into the isValidDate function as well as the date being checked?   This would at least give the isValidDate function a chance to work around this edge case.  

Something like:
```
isValid(day, 'months')
```

michelre:
@scags9876 Thanks for your reply, cause you're right. I need to check this and robustify the unit tests! I do this quickly! :)

michelre:
@scags9876 I did not really manage to reproduce your case. So I first rebased my branch against the master so I don't have conflicts.

1simonfish:
The solution @scags9876 mentioned (pass the viewMode into isValidDate) would provide a simple and robust solution. We're having this same problem and I was about to write the code for that solution.

scags9876:
@michelre After looking more into it, i think the solution you have here is sufficient to fix this issue.  Passing the viewMode into isValidDate might be a good idea for a future enhancement, but I dont think it's necessary to fix the current problem.

simeg:
@michelre this PR still needs to be rebased against master and force pushed (be careful when doing this so no commit gets lost). If you're not sure on how to do it let me know.

I am not sure what I feel about having separate methods for day, month and time. I will have to think about this.

michelre:
@simeg I've already done the rebase but I did not force pushed. I've had a lot of pain doing the rebase and I don't want to break anything

simonb-essential:
@scags9876: Can you explain then how to write an isValidDate function that disables dates before today, without disabling the entire month (if it's not the first day of the month) or the entire year (if it's not January 1)?

1simonfish:
@scags9876: Can you explain then how to write an isValidDate function that disables dates before today, without disabling the entire month (if it's not the first day of the month) or the entire year (if it's not January 1)?

scags9876:
@1simonfish Using the code in this PR, the isValidDate in example.js works to disable dates before today, without disabling the entire month.   comment out the following lines to see:

```
    viewMode: 'days',
    dateFormat: 'MMMM',
```

Using code from master, it's not possible to do what you ask, hence the need for this PR.

scags9876:
@1simonfish this function: https://github.com/YouCanBookMe/react-datetime/pull/198/files#diff-4a514f9fd4c8ed8e9fd7c71381ed1e19R10

simeg:
@michelre I'll extract your commits and push them myself if you don't feel like doing a force push, it's ok. I'll try to find some time soon to do this.

src/YearsView.js
```diff
@@ -31,16 +31,25 @@ var DateTimePickerYears = React.createClass({
         var currentYear, disabled,
             // Month and date are irrelevant here because
             // we're only really interested in the year
-            irrelevantMonth = 1,
+            irrelevantMonth = 0,
```
simeg: I am in the process of merging this commit using my own commit, but I don't understand why this change was made. Care to explain? :)

michelre: That's a good question. It seems that momentjs considers that months can be set from 0 to 11 when we use the set method. Here is the corresponding doc: http://momentjs.com/docs/#/get-set/set/.

simeg: Thanks, that makes sense.

simeg:
Hi @michelre. Thanks so much for the contribution, I just merged PR #244 where I put your changes.

Issue #202, opened 2016-11-26
View Not Updating when viewMode Changed


layneanderson:
I've tried multiple methods but I can't get the react-datetime to change the viewMode once it's initially rendered.  I have the following:

`<Datetime value={this.state.starttime} dateFormat="YYYY-MM-DD" timeFormat="HH:mm" viewMode={this.state.vmode} locale={theLocale} closeOnSelect={true} isValidDate={validDates} onChange={this.handleStartTimeChange} />`

I've got a handler on another component that changes `this.state.vmode` but react-datetime will not switch modes.  The same handler changes the value (starttime) and this does update correctly, but not the view.

PR #206, opened 2016-12-01
Fix Issue #202 - View Not Updating when viewMode Changed

layneanderson:
## Description

currentView could not be changed if the viewMode prop was updated.

## Motivation and Context

Fix Issue #202 

## Checklist

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -157,12 +157,12 @@ var Datetime = React.createClass({
 		;
 
 		if ( nextProps.value !== this.props.value ||
-            formats.datetime !== this.getFormats( this.props ).datetime ){
+            formats.datetime !== this.getFormats( this.props ).datetime ) {
             update = this.getStateFromProps( nextProps );
 		}
 
-		if ( update.open === undefined ){
-			if ( this.props.closeOnSelect && this.state.currentView !== 'time' ){
+		if ( update.open === undefined ) {
+			if ( this.props.closeOnSelect && this.state.currentView !== 'time' ) {
 				update.open = false;
 			}
 			else {
@@ -170,9 +170,9 @@ var Datetime = React.createClass({
 			}
 		}
         
-        if ( nextProps.viewMode !== this.props.viewMode ) {
-            this.setState({ currentView: nextProps.viewMode });
-        }
+        	if ( nextProps.viewMode !== this.props.viewMode ) {
+			update.currentView = nextProps.viewMode;
+        	}
 
 		this.setState( update );
 	},
--- a/DateTime.js
+++ b/DateTime.js
@@ -106,7 +106,8 @@ var Datetime = React.createClass({
 			viewDate: viewDate,
 			selectedDate: selectedDate,
 			inputValue: inputValue,
-			open: props.open
+			open: props.open,
+			currentView: props.viewMode,
 		};
 	},
 
@@ -157,7 +158,8 @@ var Datetime = React.createClass({
 		;
 
 		if ( nextProps.value !== this.props.value ||
-            formats.datetime !== this.getFormats( this.props ).datetime ){
+            formats.datetime !== this.getFormats( this.props ).datetime ||
+            nextProps.viewMode !== this.props.viewMode ){
             update = this.getStateFromProps( nextProps );
 		}
 
--- a/DateTime.js
+++ b/DateTime.js
@@ -170,9 +170,9 @@ var Datetime = React.createClass({
 			}
 		}
         
-        	if ( nextProps.viewMode !== this.props.viewMode ) {
+		if ( nextProps.viewMode !== this.props.viewMode ) {
 			update.currentView = nextProps.viewMode;
-        	}
+		}
 
 		this.setState( update );
 	},
--- a/DateTime.js
+++ b/DateTime.js
@@ -99,14 +99,14 @@ var Datetime = React.createClass({
 			inputValue = '';
 		else
 			inputValue = date || '';
-        
+
 		return {
 			updateOn: updateOn,
 			inputFormat: formats.datetime,
 			viewDate: viewDate,
 			selectedDate: selectedDate,
 			inputValue: inputValue,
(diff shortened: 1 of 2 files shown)


layneanderson:
In a use case that isn't covered by tests I found a bug in my fix.  Attempting to correct right now.

layneanderson:
This works properly now.  The issue that my first commit had was that getStateFromProps is called for user inputs like time up/down.  This had the effect of switching the currentView to what the current viewMode prop was set at.

My first commit really shows that Prop change tests should be added to the test suite (as in the Project Backlog todo's https://github.com/YouCanBookMe/react-datetime/projects/2 ).  Although a simple prop change test probably wouldn't have caught it either).  Unfortunately, I can't write this test at the moment.

layneanderson:
Hi @simeg, 

I see the label action needed.  What isn't clear though is what that action is?

Cheers,

Layne

simeg:
Hi @layneanderson, sorry for the late response. The action is my responsibility, so there's nothing you should to. Sorry if that's unclear, maybe the label needs to change. I'll look into this as soon I can find some time.

DateTime.js
```diff
@@ -99,14 +99,14 @@ var Datetime = React.createClass({
 			inputValue = '';
 		else
 			inputValue = date || '';
-
+
```
simeg: This should be removed.

example/viewModeChangeExample.js
```diff
@@ -0,0 +1,34 @@
...
+			viewMode: 'days'
+		});
+	},
+
+	componentDidMount: function() {
+		setTimeout(this.updateView, 3000);
+	},
+
+	render: function() {
+		console.log(this.state.viewMode);
```
simeg: Why is this log still here? If you think it's useful maybe do it like

`console.log('Current viewmode: ' + this.state.viewMode);`

example/viewModeChangeExample.js
```diff
@@ -0,0 +1,34 @@
...
+		return React.createElement(DateTime,
+            { viewMode: this.state.viewMode, defaultValue: moment()	 });
```
simeg: There's some empty space at the end of the line which should be removed.

ashleyyang0610:
@simeg Hi simeg! Did this solution published?

simeg:
@ashleyyang0610 no this PR is not in any published version yet. When it has gotten merged I'll publish a new version.

DateTime.js
```diff
@@ -99,14 +99,14 @@ var Datetime = React.createClass({
...
-
+        
 		return {
 			updateOn: updateOn,
 			inputFormat: formats.datetime,
 			viewDate: viewDate,
 			selectedDate: selectedDate,
 			inputValue: inputValue,
-			open: props.open
+			open: props.open,
```
layneanderson: Hi,  it was a left over from previous commit which added code there.  Linter won't catch this as its a common practice.  I'll remove it when I remove the extra spaces that were also include by the editor I am using.

example/viewModeChangeExample.js
```diff
@@ -0,0 +1,34 @@
...
+
+var Wrapper = React.createClass({
+	getInitialState: function() {
+		return {
+			viewMode: 'time'
+		};
+	},
+
+	updateView: function(format) {
+		console.log('changing state');
```
simeg: Can this be `console.log('changing viewMode to days');`? It's nice to see what the `viewMode` will change to.

example/viewModeChangeExample.js
```diff
@@ -0,0 +1,34 @@
...
+	},
+
+	componentDidMount: function() {
+		setTimeout(this.updateView, 3000);
+	},
+
+	render: function() {
+		console.log('Current viewmode: ' + this.state.viewMode);
+		return React.createElement(DateTime,
+            { viewMode: this.state.viewMode, defaultValue: moment()	});
```
simeg: This extra whitespace at the end of the line is still there, please remove it.

simeg: Also, the indentation on this line is weird. It looks like it's been adjust in the wrong direction. Sorry for these nitpicks but I want a clean code base.

DateTime.js
```diff
@@ -169,6 +169,10 @@ var Datetime = React.createClass({
 				update.open = this.state.open;
 			}
 		}
+        
+        if ( nextProps.viewMode !== this.props.viewMode ) {
```
simeg: The indentation of this whole block is not correct either. I know it looks correct in your editor, but it never looks good on GH. I'm not sure why but I will try to come up with a solution for this..

I usually copy whitespace from other parts of the file and paste, that usually works. And then double checks when I have pushed to GH.

Anyway, this should be fixed for this PR.

simeg:
@layneanderson 
>Unfortunately, I can't write this test at the moment.

Why do you say that?

DateTime.js
layneanderson: Sorry @simeg I'm not going to do anymore with the whitespace.  I fixed the issue at hand, so you can take it or leave it.

simeg:
@layneanderson 

It's ok that you don't want to fix the whitespace, I can fix that later. But I am curious to why you're saying that we can't test this, why is that?

layneanderson:
The test would be complicated* to replicate the instance where the bug was occurring.  I unfortunately don't have the time at the moment to learn the testing methods and write this myself.

I've been using this fix in production code for a while now and know it's working well.  The example could be deleted if you are comfortable with that.

*_(more complicated than I've seen in my quick purusal of the tests written so far for this component)_

simeg:
I totally forgot I can edit the files, so I am sorry about those nitpick comments.

I looked in to writing a test for this, and it turns out the current test setup does not support changing props for a component without completely creating a new one, which kinda sucks. I spent some time trying to make it work but I came to the conclusion that it should be done separately from this.

So I'm glad with this PR now and will merge it. Thanks for contributing yet again :)

Issue #174, opened 2016-10-26
onBlur event triggers only when choosing values from the calendar


ivag:
Hello,

I'm using react-datetime@2.6.0 and i have question about onBlur event:
1) I select a date from the calendar, e.g. 2016-10-10 
2)  Then i change the value to 2016-10-1 which is invalid date and i'm using onBlur event because i want to revert it to the old value: 2016-10-10.
But onBlur() triggers only when i choose dates from the calendar, not when i leave the focus of the input field. Is there any possibility to change this?

Thanks a lot!

simeg:
Hi @ivag. 

When you say "when i leave the focus of the input field", do you mean that you use the TAB key to move away from the input field? Or how do you switch focus?

ivag:
I meant 'leave the focus' by clicking outside the input file(date field) or
selecting another field. I'm not using TAB.

On Nov 8, 2016 10:43 PM, "Simon Egersand" <notifications@github.com> wrote:
>
> Hi @ivag.
>
> When you say "when i leave the focus of the input field", do you mean
that you use the TAB key to move away from the input field? Or how do you
switch focus?

simeg:
I'll look into this, it sounds weird. Thanks.

ivag:
The thing is that, i want to implement a validation in case the user enters invalid date manually (typing the date). Tell me if you need more details in order to reproduce it. Thanks.

simeg:
I'm unable to reproduce this using 2.6.0. As you can see here on [line 334](https://github.com/YouCanBookMe/react-datetime/blob/master/DateTime.js#L334) the `onBlur` method is called on `handleClickOutside`. It would help if you could reproduce your exact environment in codepen ([here's a project which you can fork](http://codepen.io/simeg/pen/mEmQmP)).

alainesse:
If the open prop is true, the onBlur is not called on the inputs onBlur

simeg:
@alainesse are you using the latest version?

alainesse:
@simeg 2.7.5

alainesse:
http://codepen.io/anon/pen/Mbmdde

simeg:
Yes, I can see it. I'll look into it. I'm guessing this is the same bug @ivag is referring to.

simeg:
@alainesse in what way do you trigger the `onBlur` of the input element when the calendar is open? By focusing it and then pressing tab?

alainesse:
@simeg clicking outside the input box.

[...]

PR #209, opened 2016-12-03
Add option for calendar to start in the open state

solomon23:
Adding the option `initializeOpen` which allows the control to be first rendered in the option state.  

## Description
Adding a new option that allows for the datetime calendar to start in the open position regardless of if there's input or not.

## Motivation and Context
This solves a problem mentioned in issue #174 where if you use the `open` prop the onblur events are then ignored.  There isn't apparently a way to have it start open and still honor the clicking outside to close it

## Checklist

- - [x] I have updated the documentation accordingly.
- - [x] I have updated the TypeScript type definition accordingly.


simeg:
Hi @solomon23 and thanks for your contribution!

To me this is a change that is well needed, but I'm not sure this is the proper way to do it. If this gets merged we would have the `open` prop as well as this `initializeOpen` prop. To me that is confusing and I'm sure other people would be confused too.

We have a backlog for V3, where we will fix the `open` prop to work as intended (when `true` the calendar will be open). Again, thanks for your contribution but this will not be merged right now. Your code might be used in the V3 release though. I hope this makes sense.

Thanks.

PR #212, opened 2016-12-05
Create readOnly prop

tmcintire:
## Description
In this commit, I added a readOnly prop to the component.  This is making use of Reacts readOnly in order to keep from showing soft keyboards on mobile devices.  

## Motivation and Context
I noticed when on mobile, it was quirky getting the input to work properly because the soft keyboard kept popping up.  I added this as an option so the keyboard won't pop up at all and it makes the field read only, but the datepicker still works.  

## Checklist
- [X ] My change required changes to the documentation.
- - [ X] I have updated the documentation accordingly.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -38,7 +38,8 @@ var Datetime = React.createClass({
 		open: TYPES.bool,
 		strictParsing: TYPES.bool,
 		closeOnSelect: TYPES.bool,
-		closeOnTab: TYPES.bool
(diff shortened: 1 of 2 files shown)


simeg:
Hi @tmcintire.

Can't this be solved using the `inputProps` property? Like `inputProps: { readOnly: true }`?

tmcintire:
Thanks for the tip.  I'm still learning

Issue #44, opened 2015-12-09
"open" prop issue


Youpinadi:
Hi,

I have a component that wraps a DateTime (there is a small button with a calendar icon at the end).

Here is the code 

``` javascript
import React, {Component, PropTypes} from 'react'
import FormInput from 'components/form/form-input/form-input.jsx'
import ReactDatetime from 'react-datetime'

export default class Datetime extends Component {
  state = {
    open: false
  }
  toggleOpen() {
    this.setState({open: !this.state.open})
  }

  render() {
    let {error,hint,className,...other } = this.props

    return (
      <FormInput className={`form_datetime form_input--merged ${this.props.className}`} error={this.props.error} hint={this.props.hint}>
        <ReactDatetime {...other} open={this.state.open}/>
        <button className="btn icon-end" onClick={::this.toggleOpen}><i className="icon icon-calendar"/></button>
      </FormInput>
    )
  }
}
```

Changing state.open doesn't do anything
I already tried a patch in `componentWillReceiveProps`:

``` javascript
if( nextProps.value != this.props.value || nextProps.open != this.props.open) {
    update = this.getStateFromProps( nextProps );
}
```

But when i do this, the calendar won't close on blur.

Basically i just want to be able to control the opening of a calendar, but the open prop doesn't do it.

Thanks in advance,

Nadir

halkeye:
what about 
```js
DateTime.prototype.componentWillReceiveProps = function (nextProps) {
  if (this.props.open !== nextProps.open) {
    this.setState({ open: nextProps.open });
  }
};
```

C3PablO:
I had the same problem and I solved it using the answer at issue #57 using `this.refs.picker.closeCalendar()`. That works but the `open` prop it's another matter. I think that is working just for the initial state

schickling:
This seems like a bug. Applying @halkeye's fix works for me.

Could you please merge this in @arqex @simeg?

TheSharpieOne:
A larger problem which affects all of the props is that they currently cannot be changed unless the `value` prop is changed.

[this](https://github.com/YouCanBookMe/react-datetime/blob/master/DateTime.js#L150-L152) only sets the update props when value has changed.

goooseman:
Maybe it is a hack, but I've coped with this issue, just passing `hidden` className (which sets display: none) to `DateTime` like this:
```
<DateTime
        input={false}
        open
        className={props.dateBoxOpened ? '' : 'hidden'} />
```

[...]

PR #216, opened 2016-12-15
Add workaround for programmatically open and close datepicker.

borisarzentar:
Workaround to programmatically open/close datepicker

## Description
Change how component checks new props so we can control open prop separately.

## Motivation and Context
This is workaround for issue #44 
This will solve problems with controlling "open" property from outside of datetime component.
Workaround assumes that you have parent container that controls datetime component. (See the example).
You can create controlled component by forwarding control "open" prop to datetime and component will react to prop change.
Also we need feedback when datetime closes itself in order to change our control value in parent component, and for this we can use onBlur function handler in parent component.
This doesn't extend existing API.

## Checklist

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -156,14 +156,24 @@ var Datetime = React.createClass({
 			update = {}
 		;
 
-		if ( nextProps.value !== this.props.value ||
-            formats.datetime !== this.getFormats( this.props ).datetime ){
-            update = this.getStateFromProps( nextProps );
+		if ( nextProps.value !== this.props.value){
+      update = this.getStateFromProps( nextProps );
+      // Close on select only makes sense if we actually changed value.
+      if ( this.props.closeOnSelect && this.state.currentView !== 'time' ){
+        update.open = false;
+      }
 		}
 
+		// Separated format update
+		if ( formats.datetime !== this.getFormats( this.props ).datetime ){
+			update.inputFormat = formats.datetime;
+		}
+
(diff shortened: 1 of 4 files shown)


simeg:
Hi @borisarzentar and thanks for your contribution. My first impression of your PR is that I do not like the word _workaround_, I prefer not having any workarounds in the codebase. 

I like the idea and effort though. On our roadmap we're currently working on a V3 of this component where we will probably change how the `open` prop is working, so to me it makes sense to not fiddle with the current stable API right now (this is kinda a breaking change).

Thanks for this, I might use some of the code/example/tests it in V3 but I'm closing this for now.

edy:
As a workaround until the next release, here's how I open the DatePicker:

```
<div className="input-group form-group-sm">
    <Datetime
        value={this.props.value}
        onChange={this.props.onChange}
        inputProps={{ref: (input) => {this.input = input;}}}
    />
    <span className="input-group-addon" onClick={() => {this.input.focus()}}><i className="fa fa-calendar"></i></span>
</div>
```

it closes himself if you click somewhere else.

simonedavico:
@edy thanks for the workaround! Any idea as to how to modify onClick to open the datepicker if it is closed and close it if it is open? I tried keeping track of the open/closed state and focusing/blurring depending on that but it didn't work unfortunately.

edy:
@simonedavico sorry, i have no idea

PR #229, opened 2017-01-01
Add support for Yarn

simeg:
Yarn support requires an autogenerated `yarn.lock file`. This commit also
lets Travis use Yarn.


simeg:
This change simply **adds the option** to install and manage the dependencies for this repo using Yarn instead of npm. If you prefer using npm you will not notice anything different.

Installing all dependencies using **npm** (nothing cached): _57.50s_ :bike:
Installing all dependencies using **Yarn** (nothing cached): _18.09s_ :car:

Issue #18, opened 2015-09-03
isValidDate should handle time as well


Youpinadi:
futureDate(current) {
        let today = moment();
        return current.isAfter(today)
    }

if i use this function as a validator, it will blur today's day, but it shouldn't.
The day should be available and only hours ans minutes in the future should be  available

iansinnott:
:+1: I've run across this issue as well. Currently we have to implement custom logic to manage time validity. Implementing an `isTimeValid` prop would be excellent.

arqex:
Sorry I could not answer this before.

It is a hard issue this one, because the UI of the datepicker makes difficult to select the valid time.

Imagine you set a function that only allow hours from 6 to 8 and 10 to 12. Pressing up and down wouldn't work ok.

That's why when a date is valid you can select any time.

If you have a solution or idea to handle this valid time issue, just leave it here.

Youpinadi:
just don't handle multiple time intervals, the most common case is just to enable timepicking after a specific datetime ;)

blackChef:
Why not show timepicker as a drop down list?

tunessofia:
@arqex are the team working on a solution for this? would be nice just to have some kind of limit, even without the time intervals

simeg:
@tunessofia from what I know no one is working on this. Your description is of `timeConstraints` is correct. 

If we would implement a `isValidTime` as @Youpinadi suggests (don't handle multiple time intervals), we wouldn't have to care about the UI stuff @arqex mentions. 

If people instead want to be able to control specific time intervals for specific days it becomes a bit trickier.

phuhl:
I have seen time-pickers that works like this:
![unbenannt](https://cloud.githubusercontent.com/assets/16835755/19597635/06d212fe-9796-11e6-9f4c-face532d30de.PNG)

You would not have the problem with illegal ranges becaus you can scroll over them.

Never the less i'd love to see a `isValidTime`- Funktion for determining a valid start-time for the picker as @tunessofia explained.

Manish-938:
@Youpinadi use isSameOrAfter to allow today date. The second argument should be day(string).

Like this: current.isSameOrAfter(new Date(), "day");

PR #271, opened 2017-02-13
Feature: tight time selection to a datetime period boudaries

vladmyr:
Problem: using `react-datetime` restrict time selection to a certain specific datetime period with millisecond precision.  

### Description
This PR extends the component's API with two new parameters - `boundaryStart` and `boundaryEnd` which set the start and end datetime period respectively. The purpose of it is to tight time setting to the datetime period.

### Motivation and Context
- The problem and solution is described above
- Request for such kind of feature was raised in #18
- If this solution gets accepted I'll provide necessary tests as well as TypeScript definitions 
- _Extra idea_: I believe with defined `boundaryStart`, `boundaryEnd`, or any of those two, `isValidDate` should be automatically generated with a possibility for user to override it

### Checklist
```
```

**UPD:** further work is required


simeg:
Hi @vladmyr, thanks for the contribution. As you probably understand this is a significant change that will take some time to review, but according to the description I think it sounds very useful.

About the naming of the properties, don't you think it would be better to just provide a single function `isValidTime` instead of two props for each boundary? I know it kinda breaks your whole implementation but I like consistency with `isValidDate`. What do you think?

vladmyr:
Hello @simeg. Thanks for the response! I'd like you to postpone the review as it turned out after some manual testing it is not production ready yet.

I understand that it is not consistent with current API and it might not match the code responsibility philosophy you are following but the idea of introducing two new props was to hide the whole implementation of `isValidTime` and `isValidDate`. The way I see it `isValidTime` will depend on time arguments as well as `currentDate`, `selectedDate` and `timeConstraints` exposing the way it will be implemented to a user/developer. I'd say it is not very convenient. In my opinion, I don't see the reason why both of approaches could not exist simultaneously. I guess firing warnings in case both APIs are used would be enough.

What are your thoughts?

simeg:
Closing due to inactivity, if you're still working on this let me know and I'll re-open it.

PR #274, opened 2017-02-16
Optimize dependencies

MIOB:
Right now building module which add react-datetime as dependency and do not add moment to dependencies cause errors like

	ERROR in ./~/moment/moment.js
	Module build failed: Error: ENOENT: no such file or directory, open 'C:\projects\internal\webrtc-demo\node_modules\moment\moment.js'
		at Error (native)
	 @ ./~/react-datetime/DateTime.js 4:10-27`

Moving moment from devDependencies to dependencies should fix this issue.
Also move react and react-dom to dependencies  because it is modules required for production.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,11 +63,8 @@
     "jest-cli": "^18.1.0",
     "jsdom": "^7.0.2",
     "mocha": "^3.2.0",
-    "moment": ">=2.16.0",
     "pre-commit": "^1.1.3",
-    "react": ">=0.13",
     "react-addons-test-utils": ">=0.13",
-    "react-dom": ">=0.13",
     "react-test-renderer": "^15.4.2",
     "through2": "^2.0.3",
     "typescript": "^2.0.10",
@@ -76,6 +73,9 @@
     "webpack-stream": "^3.2.0"
   },
   "dependencies": {
+    "moment": ">=2.16.0",
+    "react": ">=0.13",
+    "react-dom": ">=0.13",
     "object-assign": "^3.0.0",
     "react-onclickoutside": "^4.1.0"
   },


simeg:
Hi @MIOB and thanks for your contribution. It's weird that the dependencies don't get installed since they are declared as `peerDependencies`, and according to the [official node blog](https://nodejs.org/en/blog/npm/peer-dependencies/) peer dependencies should come with `react-datetime` when it is installed. 

Have you declared `react-datetime` as a dependency and then ran `npm install`? What version of `npm` are you running? (`npm --version`)

MIOB:
I have `4.2.0` npm version.
Executing `npm install --save react-datetime` cause following output:

	>npm install --save react-datetime
	webrtc@0.1.0 
	+-- UNMET PEER DEPENDENCY moment@>=2.16.0
	`-- react-datetime@2.8.6
	  +-- object-assign@3.0.0
	  `-- react-onclickoutside@4.9.0

	npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.0.0 (node_modules\chokidar\node_modules\fsevents):
	npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.0.17: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})
	npm WARN react-datetime@2.8.6 requires a peer of moment@>=2.16.0 but none was installed.

It seems that npm3 [no longer automatically download peerDependencies](http://blog.npmjs.org/post/110924823920/npm-weekly-5):

> We will also be changing the behavior of peerDependencies in npm@3. We won’t be automatically downloading the peer dependency anymore. Instead, we’ll warn you if the peer dependency isn’t already installed. This requires you to resolve peerDependency conflicts yourself, manually, but in the long run this should make it less likely that you’ll end up in a tricky spot with your packages’ dependencies

arqex:
Sorry, but peer dependencies are not meant to be downloaded by react-datetime. If somehow you ended up with 2 different react/moment versions installed at the same time you till be facing a lot of problems in your project.

Please add react, react-dom and moment to your project's dependencies.

PR #298, opened 2017-03-30
Fixed milliseconds onChange event

artemdemo:
Fixed strict comparison variables of different types.

### Description
There was bug related to milliseconds editing - strict comparison values of different types `Number === String` that always returns `false`.

### Motivation and Context
Without this bug-fix you simply can not use milliseconds.

### Checklist

- [ ] I have added tests covering my changes
- [x] All new and existing tests pass
- [ ] My changes required the documentation to be updated
- [ ] I have updated the documentation accordingly
- [ ] I have updated the TypeScript 1.8 type definitions accordingly
- [ ] I have updated the TypeScript 2.0+ type definitions accordingly

There is no need in additional changes (documentation, TypeScript definitions or something else) since it's only a bug-fix

### DIFF
--- a/src/TimeView.js
+++ b/src/TimeView.js
@@ -143,7 +143,7 @@ var DateTimePickerTime = onClickOutside( React.createClass({
 
 	updateMilli: function( e ) {
 		var milli = parseInt( e.target.value, 10 );
-		if ( milli === e.target.value && milli >= 0 && milli < 1000 ) {
+		if ( milli === milli && milli >= 0 && milli < 1000 ) {
 			this.props.setTime( 'milliseconds', milli );
 			this.setState( { milliseconds: milli } );
 		}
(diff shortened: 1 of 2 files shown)


src/TimeView.js
```diff
@@ -143,7 +143,7 @@ var DateTimePickerTime = onClickOutside( React.createClass({
 
 	updateMilli: function( e ) {
 		var milli = parseInt( e.target.value, 10 );
-		if ( milli === e.target.value && milli >= 0 && milli < 1000 ) {
+		if ( milli === milli && milli >= 0 && milli < 1000 ) {
```
simeg: `milli === milli` will always be evaluated to `true`. Not sure what this change would improve.

simeg:
Closing due to inactivity, let me know if you want it re-opened.

PR #306, opened 2017-04-10
Disable context menu in timepicker's up & down arrow

Agontuk:
Right now, right clicking on up or down arrow starts the counter while also showing the default context menu for browsers. This commit disables the context menu & makes the right click behavior identical to left click.


simeg:
Hi @Agontuk. Thanks a lot for this contribution!

I think everything looks good, but since this PR is not mergable at the moment I put it in my own PR (#459). I copied everything, even the commit subject + message. Hope this is ok by you! Thanks.

Agontuk:
Sure, no problem.

PR #327, opened 2017-05-10
Make number of weeks in "days view" fluid, fixes #279

### LINKED ISSUE #279: "Blank" lines with no days related to the current month displayed on certain months
### I'm Submitting a ...

```
[x] Bug report
[] Feature request
[] Support request
```

In DaysView, sometimes there is a full line of days belonging to the past month, which shouldn't be displayed at all since the whole week belongs to the previous month and not the current month and the "empty week" simply doesn't anyhow relate. Same applies for a full line of days belonging to the next month, which sometimes appear on Februarys ending with last day of week.

I see in the code you're using hard-coded constant 42 to build the calendar table (I get it is 4x7 weeks for minimal length of the month + 2x7 for the first and the last line) but I'd expect you to drop the "blank" lines completely. You should be checking whether the previous month ended with the last day of week (to drop the first line) or current month ended with the last day of week (drop the last line).

### Steps to Reproduce
1. Open the example page
2. Set date to October 2000 - the first line only contains days from previous month - nothing related to current month
3. Set date to February 1999 - the last line only contains days from previous month - nothing related to current month
4. Set date to February 1998 - both the first are not related

### Expected Results
The lines containing no days related to current month should be dropped. Of course I get you'll lose the "fixed grid" of 7x6 to 7x5 or even 7x4 but the real question here is whether having it fixed is a necessary feature. Checking other date pickers on the market (mainly via flight companies' sites) not showing unrelated lines of days is an expected behaviour.

lopis:
### Description
Currently, the number of weeks in "days view" is fixed to 5 which causes most months to either have a whole week from the previous month or from the following month being rendered. This is unusual in most calendar UIs.
This change checks if the previous and current months end on the last day of the week (depends on the locale set in momentJS) to decide if an extra week is necessary.
This also fixes tests to take this change into account.

So this:
image

Becomes this:
image

And this:
image

Becomes this:
image

### Motivation and Context
Fixes #279 

### Checklist

[x] I have added (changed) tests covering my changes

_Please let me know if this is something you would like to see merged. It would be useful to me. I have some more time I can pour into this if you think I should parametrize this instead because this has the side effect of changing the widget height._

### DIFF
--- a/tests/datetime.spec.js
+++ b/tests/datetime.spec.js
@@ -234,14 +234,16 @@ describe('Datetime', () => {
 	});
 
 	it('selected day persists (in UI) when navigating to prev month', () => {
-		const date = new Date(2000, 0, 3, 2, 2, 2, 2),
+		const date = new Date(2000, 0, 1, 2, 2, 2, 2),
 			component = utils.createDatetime({ viewMode: 'days', defaultValue: date });
 
 		utils.openDatepicker(component);
-		expect(utils.getNthDay(component, 8).hasClass('rdtActive')).toBeTruthy();
+		expect(utils.getNthDay(component, 6).hasClass('rdtActive')).toBeTruthy();
+		let dayNumber = utils.getNthDay(component, 6).text();
 		// Go to previous month
 		utils.clickOnElement(component.find('.rdtDays .rdtPrev span'));
-		expect(utils.getNthDay(component, 36).hasClass('rdtActive')).toBeTruthy();
+
+		expect(utils.getActiveDay(component).text() === dayNumber).toBeTruthy();
 	});
 
 	it('sets CSS class on today date', () => {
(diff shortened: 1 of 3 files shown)


tests/datetime.spec.js
```diff
@@ -234,14 +234,16 @@ describe('Datetime', () => {
 	});
 
 	it('selected day persists (in UI) when navigating to prev month', () => {
-		const date = new Date(2000, 0, 3, 2, 2, 2, 2),
```
lopis: After my changes , the 3rd of January no longer show in the previous month.

tests/datetime.spec.js
```diff
@@ -234,14 +234,16 @@ describe('Datetime', () => {
...
+		const date = new Date(2000, 0, 1, 2, 2, 2, 2),
 			component = utils.createDatetime({ viewMode: 'days', defaultValue: date });
 
 		utils.openDatepicker(component);
-		expect(utils.getNthDay(component, 8).hasClass('rdtActive')).toBeTruthy();
+		expect(utils.getNthDay(component, 6).hasClass('rdtActive')).toBeTruthy();
+		let dayNumber = utils.getNthDay(component, 6).text();
 		// Go to previous month
 		utils.clickOnElement(component.find('.rdtDays .rdtPrev span'));
-		expect(utils.getNthDay(component, 36).hasClass('rdtActive')).toBeTruthy();
```
lopis: After my changes, the position of the previous month would not be fixed so I changed the test to make sure the day number is the same instead.

lopis:
Closed because of conflicts.

PR #329, opened 2017-05-12
Add support for displayTimeZone prop

martingordon:
### Description
This PR adds support for a `displayTimeZone` prop that displays the date in the input in the specified time zone.

### Motivation and Context
The current options to the component only allow showing the date in either the user's local time zone or UTC. We have a use case for allowing users to specify a date outside of either of these two time zones.

### Checklist
```
[ x ] My changes required the documentation to be updated
  [ x ] I have updated the documentation accordingly
  [ x ] I have updated the TypeScript 1.8 type definitions accordingly
  [ x ] I have updated the TypeScript 2.0+ type definitions accordingly
```


martingordon:
@holyman2k I just fixed the merge conflicts. Any chance we can get this merged in soon?

tests/datetime.spec.js
```diff
@@ -813,6 +814,34 @@ describe('Datetime', () => {
 				});
 			});
 
+			it('displayTimeZone -> value should change format (undefined->America/New_York)', () => {
+				const date = new Date(2000, 0, 15, 2, 2, 2, 2),
+					momentDate = moment(date),
+					component = utils.createDatetime({ value: momentDate }),
+					displayTimeZone = (moment.tz.guess() === 'America/New_York' ? 'America/Los_Angeles' : 'America/New_York');
```
simeg: Why making this random? I don't see the benefit, and it makes the test more complex.

README.md
```diff
@@ -43,6 +43,7 @@ render: function() {
 | **open** | `boolean` | `null` | Whether to open or close the picker. If not set react-datetime will open the datepicker on input focus and close it on click outside. |
 | **locale** | `string` | `null` | Manually set the locale for the react-datetime instance. Moment.js locale needs to be loaded to be used, see [i18n docs](#i18n).
 | **utc** | `boolean` | `false` | When true, input time values will be interpreted as UTC (Zulu time) by Moment.js. Otherwise they will default to the user's local timezone.
+| **displayTImeZone** | `string` | `null` | When specified, input time values will be displayed in the given time zone. Otherwise they will default to the user's local timezone (unless `utc` specified).
```
simeg: Normally I wouldn't make this comment but since I had more comments I will. You have a typo :) `displayTImeZone` -> `displayTimeZone`

tests/datetime.spec.js
```diff
@@ -813,6 +814,34 @@ describe('Datetime', () => {
...
+
+				const valueBefore = utils.getInputValue(component);
+				component.setProps({ displayTimeZone: displayTimeZone }, () => {
```
simeg: I don't trust assertions inside these callbacks, they don't seem to work sometime. Did you verify with something like `expect(true).toEqual(false)`?

simeg:
Hi @martingordon. Sorry for the delay, but I looked at your PR and it looks good. I left three comments (some of them are marked as outdated though).

hieuhlc:
Hi @martingordon I see your interesting idea, is there any chance you could continue the work so that the PR will be merged? Really appreciate that!

hieuhlc:
Instead of `displayTimeZone` being `string|null`, how about `formatDisplay` receiving a function that converts the value into string for display purpose only
```javascript
function formatDisplay(value <moment Object>) <string> {...}
```
In that case we can pass `moment.tz` or our own custom format function. More flexible I think.

kiran-rg:
:+1: I have a similar use case. It would be great if this PR gets merged. Is there anything pending in this PR that I could help with.

package.json
```diff
@@ -40,6 +40,7 @@
   "license": "MIT",
   "peerDependencies": {
     "moment": ">=2.16.0",
+    "moment-timezone": "^0.5.13",
```
ljharb: This caused #596 - adding a peer dep is semver-major, a breaking change.

PR #335, opened 2017-05-21
Fix displayName <Unknown /> in react dev tools

dragfire:
When inspecting the component in React Dev Tools, it is displayed as `<Unknown />` Component, which is kind of annoying.
So this PR fixes this issue.

### Description
displayName is added

### Motivation and Context
Displaying `<Unknown />` instead of `<DateTime />` doesn't look good.

### Checklist

```
```


simeg:
Hi @dragfire and sorry for the delay. Your single commit has the changes from building this project, that is not necessary and I would like for you to remove it.

Can you confirm that this is still a problem? If so we could make this mergable and merge it.

simeg:
Thanks for this @dragfire. I'll push this change in a separate PR, due to the inability to merge this PR.

PR #345, opened 2017-06-02
Add datetimeFormat override for displaying value

andjosh:
I've found a use-case where we want the picker to function the same but
display the selected date separately when closed. This allows a final
override for the value displayed in the text input.

### Checklist
```
```


DateTime.js
```diff
@@ -133,10 +136,12 @@ var Datetime = createClass({
 			formats.time = locale.longDateFormat('LT');
 		}
 
-		formats.datetime = formats.date && formats.time ?
-			formats.date + ' ' + formats.time :
-			formats.date || formats.time
-		;
+                if ( !formats.datetime  || formats.datetime === true ) {
```
simeg: I find this to be weird, we are checking for either `falsey` (which it is by default) or `true`. When would this entire expression not be `true`?

DateTime.js
```diff
@@ -133,10 +136,12 @@ var Datetime = createClass({
...
+                        formats.datetime = formats.date && formats.time ?
```
simeg: Could this be a separate function? It looks very confusing. Or maybe pull out time and date into variables?

DateTime.js
```diff
@@ -41,6 +42,7 @@ var Datetime = createClass({
 			onFocus: nof,
 			onBlur: nof,
 			onChange: nof,
+                        datetimeFormat: false,
```
simeg: I'm not sure I like this naming, to me the format we are talking about is for the _value_ in the input field. Datetime is such a broad word. Maybe `valueFormat`, what do you think?

tests/datetime.spec.js
```diff
@@ -285,6 +285,22 @@ describe('Datetime', () => {
...
+				component = utils.createDatetime({ value: date, datetimeFormat: 'M&D&h' });
+			expect(utils.getInputValue(component)).toEqual(mDate.format('M&D&h'));
+		});
+
+		it('datetimeFormat=true', () => {
+			const date = new Date(2000, 0, 15, 2, 2, 2, 2),
+				mDate = moment(date),
+				component = utils.createDatetime({ value: date, datetimeFormat: true });
+			expect(utils.getInputValue(component)).toEqual(mDate.format('L LT'));
+			// Make sure time view is active
```
simeg: This comment is not true. You are actually checking for the component to _not_ have the time view active on the line below.

DateTime.js
```diff
@@ -5,11 +5,12 @@ var assign = require('object-assign'),
         createClass = require('create-react-class'),
 	moment = require('moment'),
 	React = require('react'),
+        onClickOutside = require('react-onclickoutside'),
```
simeg: I am not sure I understand the implications of this change, and why it is necessary. Could you try to explain it further? Also this sounds like a major version update because (if I understand it correctly) it would break all current components, correct?

simeg:
Hi @andjosh, sorry for the delay. Good structured PR, left you some comments.

tests/datetime.spec.js
andjosh: so true :) Will change

DateTime.js
andjosh: Yeah, very confusing. I will extract this.

DateTime.js
andjosh: Previously, the click handler was registered on the faux date input display, which caused the `onClickOutside` event to fire when clicking on the input itself. This change wraps the whole bundle in one click listener.

It potentially changes the behavior of the inputs, as it now wraps them all in the same click listener. I believe that the previous implementation was buggy, as the `onClickOutside` was actually being triggered on the first click _inside_.

PR #347, opened 2017-06-05
update state when property open is changed

jaroslav-kubicek:
### Motivation for change

Thank you for creating such useful component. There is only one small thing that bugs me:

I want to control whether calendar is open or not from the outside via property. The problem is that as long as the new instance of the component is created, boolean `open` is set in state accordingly to the prop, but if the value is changed afterwards, it's not reflected, which cause confusion. So it might be convenient to update state in such cases.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -192,6 +192,10 @@ var Datetime = createClass({
 				}
 			}
 		}
+		
+		if ( nextProps.open !== this.props.open ) {
+			updatedState.open = nextProps.open;
+		}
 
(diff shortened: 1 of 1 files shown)


simeg:
Hi @jaroslav-kubicek, sorry for the delay.

This bug was annoying to a lot of people so I believe someone else make a similar PR that was merged, so this should now be fixed. Please try a more recent version and let me know if the problem persists. I'll close this for now, feel free to re-open.

PR #349, opened 2017-06-07
Added props for tables

Vallefor:
Added props for different tables

### Description

Add props:
timeTableProps - used in time table
daysTableProps - used in days table
monthsTableProps - used in months table
yearsTableProps - used in years table
headerTableProps - used in years and months header table

### Motivation and Context

Sometimes you need to change table props. For example, some css framework adds default styles to elements (like table, ul, li and etc) but you can prevent it by adding some class to it.

### Checklist

```
[v] All new and existing tests pass
[v] My changes required the documentation to be updated
```


simeg:
Hi @Vallefor. I like the idea of being able to add css classes to each of the tables. However your PR is very confusing to me, and I'm not so keen on the naming of the props.

For a start, could you remove all the changes for the dist files? There's no need to add that when you make a PR. Thank you, and sorry for the delay.

PR #353, opened 2017-06-20
Remove React.DOM in favor of React.createElement

gwing33:
React v15.6 throws a deprecation warning about `React.DOM` factory.

### Description
In order to prevent this warning from happening, I replaced the `DOM[type](...)` with `React.createElement(type, ...)`.

### Motivation and Context
No one likes warnings from libraries.

### Checklist
- [x] All existing tests pass


layneanderson:
I'll be coming on board as a maintainer shortly.  This is likely the 1st PR I'll merge.  We are just currently waiting for the owner of the repository to come back from vacation to grant me rights.

simeg:
Hi @gwing33. Thanks for your contribution. I'm sorry the response was so late. 

Since the owner is on vacation I am merging this instead of @layneanderson. We're releasing 2.8.11 soon and this will be part of that release.

PR #362, opened 2017-07-08
Dynamic time constraints

jankarres:
### Description
Previously: You pass timeConstraints into the component, which was set at componentWillMount, with the consequence that if timeConstraints props changes, the component will user the old timeConstraints.

_Example_

Max time: 00:14:15:00
Value on mount: 00:00:00:00

To set e.g. 00:12:59:00 the timeConstraints have to look like:

```
{
    hours: {
        min: 0,
        max: 0,
        step: 1
    },
    minutes: {
        min: 0,
        max: 14,
        step: 1
    },
    seconds: {
        min: 0,
        max: 59,
        step: 1
    },
    milliseconds: {
        min: 0,
        max: 999,
        step: 1
    }
}
```

If the user changed to e.g. 00:14:??:00 (? = some value), the timeConstraints have to look like (max of seconds changed):

```
{
    hours: {
        min: 0,
        max: 0,
        step: 1
    },
    minutes: {
        min: 0,
        max: 14,
        step: 1
    },
    seconds: {
        min: 0,
        max: 15,
        step: 1
    },
    milliseconds: {
        min: 0,
        max: 999,
        step: 1
    }
}
```

I wrote the code, which refresh the timeConstraints on every componentWillReceiveProps and check for changed constraint fails at calculateState and in case set min or max value.

### Checklist
```
[x] I have added tests covering my changes => according to me knowledge no new tests required
```

### DIFF
--- a/src/TimeView.js
+++ b/src/TimeView.js
@@ -37,6 +37,46 @@ var DateTimePickerTime = onClickOutside( createClass({
 			}
 		}
 
+        if (props.timeConstraints['hours']) {
+            if (parseInt(date.get('hour'), 10) < props.timeConstraints['hours'].min) {
+                date.hour(props.timeConstraints['hours'].min);
+            }
+
+            if (parseInt(date.get('hour'), 10) > props.timeConstraints['hours'].max) {
+                date.hour(props.timeConstraints['hours'].max);
+            }
+        }
+
+        if (props.timeConstraints['minutes']) {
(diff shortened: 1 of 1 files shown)


simeg:
Hi @jankarres, thanks for contributing and sorry for the delay.

Regarding testing, you could test that if you update the component the updated props are read accordingly. There are multiple similar tests.

Issue #331, opened 2017-05-17
Refs Must Have Owner Warning


und0ck3d:
[X] Bug report
[ ] Feature request
[ ] Support request

### Steps to Reproduce
- Install the latest `react-datetime` version;
- Require and use it in any React component;

### Expected Results
Expected that component worked normally.

### Actual Results
When the previous steps are done, an error shows in the console:

```
Uncaught Error: addComponentAsRefTo(...): Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner).
```

### Minimal Reproduction of the Problem
1. I'm creating a component to be reused in other projects via installing with NPM. In that component I'm using `react-datetime`.
2. In the main project, I require this new component (mine) and `react-datetime` is provoking the error mentioned before (because if I don't use this package it doesn't trigger the error).

**The following code is redacted for simplicity**

My Component:

```javascript
import React from 'react';
import PropTypes from 'prop-types';
import DateTime from 'react-datetime';

class AvailForm extends React.Component {
  render() {
    let { date } = this.state;

    return (
      <div>
        <DateTime date={ date } onChange={ this._onChangeDate } />
      </div>
    );
  }
}

export default AvailForm;
```

On the main project I do:

```javascript
import React from 'react';
import AvailFrom from 'availform'; // not actual package name, just for eg. purposes

export default class SomePage extends React.Component {
  render() {
    return (
      <AvailForm />
    );
  }
}
```

Finally, on main project's main file I'm rendering the component like (simplified):

```javascript
import { render } from 'react-dom';
import SomePage from './containers/SomePage';

render(<SomePage />, document.getElementById('some-page'));
```

Please, let me know if I haven't been clear in any step or if further information is needed.

Thanks everyone beforehand for your time and help!

joaocrq:
I've been reading around and it seems that the problem was caused because in the main app I had React as a dependency and in my component too, which was causing to having multiple instances of React running. I "fixed" it by setting my component's `react` and `react-dom` dependencies as peerDependencies, however now I can't build the component with Webpack...

I'm not sure if this discussion should continue here, but can anyone help me out please?

bruno-gouveia:
I think the problem comes from react-onclickoutisde dependency, the version 6.1 fixes this.

https://github.com/Pomax/react-onclickoutside

exah:
[...]

PR #364, opened 2017-07-11
Update react-onclickoutside to 6.1.0

john-osullivan:
Fixes https://github.com/YouCanBookMe/react-datetime/issues/331 , which is a total breaking bug.

### Description
Pretty minimal change, just version bumping a dependency to get around an issue which `react-datetime` is imported into another component which is itself being imported.  All of the tests which were already passing continued to pass with the new version.

### Motivation and Context
PR should be merged because it fixes an open issue.

### Checklist
```
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "@types/react": ">=15",
     "object-assign": "^3.0.0",
     "prop-types": "^15.5.7",
-    "react-onclickoutside": "^5.9.0"
+    "react-onclickoutside": "^6.1.0"
   },
   "pre-commit": [
     "lint",


john-osullivan:
Well that's odd, the Travis build failed with a totally different set of errors than I experienced on my local machine.  Testing further.

john-osullivan:
> Update 2: The plot thickens: `react-onclickoutside@6.1.0` no longer supports `create-react-class`.  They're [surprisingly opinionated](https://github.com/Pomax/react-onclickoutside#which-version-do-i-need-for-which-version-of-react) about which of their versions matches the proper react versions.  A good amount of this library uses mixins, so migrating to pure classes doesn't seem to be straightforward.  Version `5.11.x` is explicitly designed to depend on `create-react-class`, but it still exhibits the issue shown here.  Looks like there isn't a way to fix that issue without doing a non-trivial refactor such that react-datetime stops using mixins.

PR #383, opened 2017-07-27
Add renderInput prop

### LINKED ISSUE #153: Enable pre and post content to be added

danfo:
### Description

Added an optional `renderInput` prop which replaces the rendering of the input element. The accepted function has `openCalendar` (a function which opens the calendar) and the default calculated `props` for the input. Must return a React component or `null`.

### Motivation and Context

  * Allows further customization of what is rendered (e.g. adding an icon button to open the calendar) without any breaking changes
  * Resolves #153 and related issues

### Checklist

```
```


heatherbooker:
It would be awesome to pass closeCalendar as well - seems to just make sense to have them both. Is this something that could be done or that you would be interested in seeing a modified PR for?

goooseman:
I'm using the danfo's fork for now. Will this PR be merged some day? Do you need any help to do smth to make this happen?

simeg:
Hi @danfo. Sorry for the delay, I just looked through this and the code looks great. It's not breaking any backward compatibility which is critical. Well structured PR 👌, thanks a lot.

Since this branch lives on your fork I can't push any changes to it, so I created a new PR which I can control. Will close this PR and continue in #464.

PR #396, opened 2017-08-08
add manualEditable to props

lordisback:
### Description

I add a manualEditable props to DateTime in order to controll the input element 
### Motivation and Context

When the input props set to be true, the entire input element wil be removed from the DOM, but there are some cases that we only want to make the input element not manully editable, thats why I add the maualEditable props to DateTime component
### Checklist

```
[x ] I have added tests covering my changes
[x ] All new and existing tests pass
[x ] My changes required the documentation to be updated
  [x ] I have updated the documentation accordingly
```


simeg:
Hi @lordisback.

What this PR is trying to achieve is possible using the `inputProps` property, which you can find more about here: https://github.com/YouCanBookMe/react-datetime#api

PR #403, opened 2017-08-15
Add onViewDateChange callback when the user changes the currently viewed date

InvisibleBacon:
### Description
I added a callback that is triggered whenever the internal `viewDate` state value changes.

### Motivation and Context
This is useful if you need to fetch data from the server in response to the user changing their view. In our application we wanted to color-code the days in the month view, and we wanted to know when the month was changed so we could fetch the coloring data from the server for the newly displayed month.

### Checklist

```
```

NOTES:

- I didn't see the TypeScript 2.0+ typeings file in the typings directory. So I didn't update those. I might be missing something.
- A few of the existing tests didn't pass on my machine because I was running in EST. I made sure my new tests will pass in any timezone, but I didn't fix the existing tests.


src/MonthsView.js
```diff
@@ -82,7 +82,7 @@ var DateTimePickerMonths = onClickOutside( createClass({
 	},
 
 	renderMonth: function( props, month ) {
-		var localMoment = this.props.viewDate;
+		var localMoment = this.props.viewDate.clone();
```
simeg: Nice catch. This might be the source of some test problems we've been having.

simeg: I tried it and unfortunately no :(

simeg:
Hi @InvisibleBacon, and thanks for contributing. Sorry for the delay.

I tried to get it to a mergable state, then I remembered that this functionality has already been implemented in an upstream PR. See `onViewModeChange` in [the API](https://github.com/YouCanBookMe/react-datetime#api).

InvisibleBacon:
Thanks @simeg. So are you saying that `onViewModeChange` is triggered when switching to different months/years/etc.?

I am also seeing `onNavigateBack` and `onNavigateForward` that look like they were intended for this purpose. Is that what you were referring to?

simeg:
Oh, no I think I misunderstood your PR then. `onViewModeChange` will not trigger on any date change.

Sorry, will look into this PR ASAP.

arqex:
Sorry for getting late into this PR but thanks! Many changes happened since it was created and now we have the `onNavigate` to detect the changes of views. Closing

Issue #401, opened 2017-08-14
Add a defaultTime prop


shawninder:
### I'm Submitting a ...

```
[ ] Bug report
[x] Feature request
[ ] Support request
```

### Steps to Reproduce
Add a `defaultTime` prop to the element

```
<ReactDateTime
  dateFormat="YYYY-MM-DD"
  timeFormat="HH:mm:ss"
  defaultTime="23:59:59"
/>
```

### Expected Results
When clicking on a day in the calendar, if `defaultTime` is set, the datetime obtained should show the provided default time (in my example above: "23:59:59") instead of "00:00:00" as it currently does

### Actual Results
Nothing, hence the feature request

### Why?
I am using this component to make a DateTime **range** picker and I'd like the default times to be "00:00:00" for the "From" input, but "23:59:59" for the "To" input, yet I need the `defaultValue` to be `undefined` since I want to support open-ended ranges. Such a new feature would make this use-case really easy:

```
<label>FROM:</label>
<ReactDateTime
  dateFormat="YYYY-MM-DD"
  timeFormat="HH:mm:ss"
/>

<label>TO</label>
<ReactDateTime
  dateFormat="YYYY-MM-DD"
  timeFormat="HH:mm:ss"
  defaultTime="23:59:59"
/>
```

kimooz:
+1

shawninder:
I made a PR for this: https://github.com/YouCanBookMe/react-datetime/pull/404

rsl:
i'd love to see this implemented. defaultValue doesn't fit my needs either.

lenogen:
I would love this as well.

omkar-joshi:
+1

dep:
+100

lchangdev:
+1

michaeldybal:
+1 Same case with date range

Amjad-mozanta:
+1

vebaspect:
+1

Jenny4831:
+1

giorgi-m:
any news on this?

sneas:
`initialViewDate` property could be used to achieve the desired result.

CarlosPhilipe:
+1

PR #404, opened 2017-08-18
Added `defaultTime` prop

shawninder:
### Description
The DateTime component now accepts an optional `defaultTime` prop which specifies which time to set when selecting a day.

### Motivation and Context
This is particularly useful if you're using two instances of the component to make a date-time **range** picker. In this case the *from* datetime picker can stay as-is, but the *to* datetime picker should default to the end of the day, not the beginning, so that you can just select a *from* day and a *to* day, and both will be included in the range.

### Checklist
I didn't finish this checklist, but I can finish it if there is any interest in merging this. Otherwise, this suits my needs so I would just leave it. If anyone else want to finish this checklist and send me a PR, that would be awesome.

```
### Additional notes
When cloning the original component (YouCanBookMe/react-datetime), a lot of tests didn't pass. I have not fixed them so they still fail here, but all the tests that passed before these changes still pass.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -13,6 +13,7 @@ var Datetime = createClass({
 	propTypes: {
 		// value: TYPES.object | TYPES.string,
 		// defaultValue: TYPES.object | TYPES.string,
+		// defaultTime: TYPES.string,
 		onFocus: TYPES.func,
 		onBlur: TYPES.func,
 		onChange: TYPES.func,
@@ -37,6 +38,7 @@ var Datetime = createClass({
 		return {
 			className: '',
 			defaultValue: '',
+			defaultTime: '',
 			inputProps: {},
 			input: true,
 			onFocus: nof,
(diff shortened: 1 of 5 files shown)


shawninder:
This new feature was suggested here: https://github.com/YouCanBookMe/react-datetime/issues/401

shawninder:
I am publishing this to a private npm repo, so don't forget to revert the name in the package before merging this PR

simeg:
Hi @shawninder. Thanks for your contribution.

I think this would be a good feature to have. Although I don't want to merge this PR since it's really the first commit that is interesting, so I might just take your code and make my own PR. Hope that's ok with you. Thanks again for this nice feature. Keeping this PR open until someone fixes this.

shawninder:
This has indeed become my own fork for my own uses, I apologize. If you make your own PR here with only that commit, please link to it here and I can deprecate/remove mine.

Another thing to consider is that this new feature has only been tested in the specific configuration in which I use react-datetime. Your mileage may vary.

Cheers!

et304383:
Can we get some traction on this?  It's absurd that we cannot set this by now.

When a user picks a start and end date of the same date, it's assumed that it will filter based on a single day, but now with default behaviour it's a search of date between a start and end date that are identical, which means no results.  No one would ever want this as a desired behaviour.

arqex:

Issue #408, opened 2017-08-24
Typescript - Installing with npm v4+ causes duplicate react types.


cwmacdon:
### I'm Submitting a ...
```
[x] Bug report
[ ] Feature request
[ ] Support request
```

### Steps to Reproduce
1) Use `npm v4.x` (I've tried with `4.3.0` and `4.6.1`) (not reproducible with `3.10.10`)
2) Have a `package.json` that has:
```
{
  "scripts": {
    "build": "tsc index.tsx"
  },
  "dependencies": {
      "react-datetime": "2.9.0",
      "react": "15.6.1",
      "react-dom": "15.6.1",
      "moment": "2.18.1"
  },
  "devDependencies": {
    "@types/react": "15.6.1",
    "typescript": "2.4.2"
  }
}
```

### Expected Results
My code to compile.

### Actual Results
A bunch of errors about react types not matching due to `react-datetime` pulling in `@types/react@16.0.5`

### Minimal Reproduction of the Problem
Using the `package.json` above, and the following `index.tsx`:
```
import * as React from 'react';
import * as Datetime from 'react-datetime';
```
1) `npm i`
2) `npm run build`

You get a few pages of errors that look like this:
`node_modules/@types/react/index.d.ts(3400,19): error TS2320: Interface 'Element' cannot simultaneously extend types 'ReactElement<any>' and 'ReactElement<any>'.
  Named property 'type' of types 'ReactElement<any>' and 'ReactElement<any>' are not identical`

### Other Information (e.g. stacktraces, related issues, suggestions how to fix)
Works fine with npm v3 since v3 does not install a separate version of `@types/react` in  `/node_modules/react-datetime/node_modules/@types/react`

cwmacdon:
I'm fairly certain (but in the process of verifying) that this is due to the PR which moved `@types/react` from `devDependencies` to `dependencies`.

cwmacdon:
I can confirm that this is indeed the cause.

simeg:
See my questions in #409.

simeg:
Closing as PR #409 was merged.

PR #409, opened 2017-08-24
Move @types/react back to devDependencies

cwmacdon:
### Description
See https://github.com/YouCanBookMe/react-datetime/issues/408 for details

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "react-dom": ">=0.13"
   },
   "devDependencies": {
+    "@types/react": ">=15",
     "babel-core": "^6.22.1",
     "babel-jest": "^18.0.0",
     "babel-loader": "^6.2.1",
@@ -77,7 +78,6 @@
   },
   "dependencies": {
     "create-react-class": "^15.5.2",
-    "@types/react": ">=15",
     "object-assign": "^3.0.0",
     "prop-types": "^15.5.7",
     "react-onclickoutside": "^5.9.0"


simeg:
Hi @cwmacdon and thanks for your contribution! As I'm not a TypeScript user I'm having a hard time seeing the impact of this change. With this change we're basically saying that the package `@types/react` is not a requirement to use this component, just during development. I'm not sure what `@types/react` brings to the table in the first place. If you know, would you mind explaining it to me? Thanks.

cwmacdon:
Hey @simeg,

Yes, all TypeScript definition files (any module in the @types namespace) should be development dependencies only. See [this TypeScript issue](https://github.com/Microsoft/TypeScript/issues/9731#issuecomment-232864820)

---- Some details below on what is actually going on ----

The issue here is that npm is resolving the packages and gives a special priority to packages defined as `dependencies` vs `devDependencies`.

Here is a detailed look at how things are being resolved and what is going wrong:

If you look at the example `package.json` file from the [issue](https://github.com/YouCanBookMe/react-datetime/issues/408), you can see how a typical typescript project should be set up. If my project is being developed against react v15 and I include your `react-datetime` as a dependency, then npm will resolve as follows:

1) looks at `react-datetime` dependencies, and installs `@types/react@16.0.5` in top level `node_modules` (or latest >=15)
2) looks at local devDependencies, and sees `@types/react@15.6.1`. It will move `@types/react@16.0.5` to `node_modules/react-datetime/node_modules/@types/react`, and then install `15.6.1` in my top level `node_modules`

Now the `node_modules` looks like this:

- react
- react-datetime
  - Datetime.d.ts
  - node_modules
    - @types
      - react
- @types
  - react@15.6.1

Now when the TypeScript compiler is looking at the following code:
```
import * as React from 'react';
import * as Datetime from 'react-datetime';
```

It grabs `@types/react@15.6.1` from the first line.
Then from the second line it grabs: `react-datetime/Datetime.d.ts`.
This file imports from `react`, and the compiler first attempts to resolve locally, so it looks inside `react-datetime/node_modules/@types` and pulls in `@types/react@16.0.5`
Now the compiler has multiple competing definitions, and can't know which one is right, so it throws errors.

cwmacdon:
If you run `npm dedupe`, it figures out that the `@types/react@15.6.1` solves the constraint defined by `react-datetime` looking for `@types/react >= 15`, so it removes the extra package, and everything works fine. Unfortunately this isn't a reasonable workaround, since dedupe takes forever, and can't be targeted at a single package.

simeg:
Thanks for that thorough explanation. This seems to be a reasonable change. Thanks a lot for your contribution.

PR #411, opened 2017-08-27
Fix build files

layneanderson:
Fix build files.


Simek:
After this change "react-datetime" stops working on my production environment. Every browser that I've tested freezes when I tried to enter a page containing date picker. No console log/warn/err in browser, no build errors, reverting to 2.9.0 immediately fixes those issues.

PR #426, opened 2017-09-22
Fix URL for 'uncontrolled elements'

Grahack:


simeg:
Hey @Grahack, thanks for this!

PR #435, opened 2017-09-30
InitialRenderDate a replacement for IsValidDate checking

JesseChrestler:
### Description
This is to replace the need for the code to check if the date rendered is in the validate date range. InitialRenderDate will allow the user to specify the date at which the calendar will render on the initial load. This includes both future and past dates. 

### Motivation and Context
The original fix was a bit dangerous and had bugs. This way is faster, and cleaner and can be used outside of the context of isValidDate checking

### Checklist
```
```


simeg:
Hi @JesseChrestler.

This pr has +22,193 −22,166, looking into the changes it looks like you've automatically linted every single file. Please fix this, otherwise it's hard to review your changes.

arqex:

PR #437, opened 2017-10-05
WIP - Refactor src into ES6 and fix misc bugs, tests and styles for v3 (need feedback)

bjbrewster:
### Description

Hello, 

I would like to get the ball rolling on v3 and bringing this project into 2017.

I am having fun rewriting the src files in ES2015 to make it easier to maintain and have also cleaned up a lot of bad code! I've also removed lots of unused styles and fixed a few bugs.  I tried using ES6 classes but found the intermediate libs and bundled packages were very bloated so I reverted to using create-react-class and was amazed at how much smaller it is. My min bundle is about 30k compared to the current 40k. :-)

I am using babel cli to convert the /src files to /lib files in plain js (as it is now). The package.json "main" entry now points to /lib/DateTime.js so it should behave the same as it is now.  I've added a new package.json "module" entry point to /src/DateTime.js which is used by Webpack and other bundlers looking for modern js.

I've left the interface almost exactly the same except renderDay, renderMonth, and renderYear are now render components - either a react class or a stateless react render method that accepts a props argument. The default render methods were basically stateless react render methods anyway, except they received date/month/year/selectedDate as extra args. Now these args are in the main props. This allows the custom render methods to add context props if required. Or use full React component classes if needed.

I'm investigating using Fecha instead of moment js as it is a much smaller Date library based on moment (only 2k min gzipped).

Anyhow, I'd appreciate some feedback and would like to discuss a few things:

1. Is it intentional that the Date Picker Views are rendered even when the Picker is not "open"? At the moment it's using a CSS class to hide/show the component which is not the React way and IMO really bad. e.g. If you have a form with a list of 1000 items, each with their own DateTime input box, they will all have a rendered DateTime Picker view as well (just hidden using CSS). This is very inefficient. It should just not include the Picker view when not open. I've made my v3 branch do this and updated many tests accordingly. If the desire for it to be always open so users can add CSS transition animations, we add CSS classes for on enter and about to leave and delay the removal to allow for this.

2. I see someone updated the react-onclickoutside dependency to the latest ^6 version but it has peer deps of React 15.5.x+ or 16.0+ which is a problem unless we update this project's peer deps to the same. For now, I've updated my v3 branch to use latest react-onclickoutside@^5 that isn't as restrictive.

3. The current TimeConstraints seem completely broken to me. Why would you even want to restrict mins or seconds let alone hours with the same values for every day? And a min/max mins or secs seems to only make sense as a whole min/max time (hours, mins and secs together), not separate as it is now. A lot of people are complaining that this feature is completely useless and I can see why. For now I've disabled custom time constraints in my v3 branch as the current feature is broken IMO. A better solution would be a callback that is called with the selected day that returns an array of allowed min/max time ranges. An example usage would be for a booking system where there are slots free at 8:30am to 10:15am and 3:45pm to 6:15pm. I see someone is working on a better time input (like using an analogue clock) so I'll leave it to others to decide a better design.

Feedback is more than welcome.

### Checklist

```
```


simeg:
Hi @bjbrewster. 

This sounds great, I like that you're doing this. Why did you close it though?

1) I'm not sure but I don't think this is intentional. I'm not sure what the _React way_ is but I'd be up for changing that.
2) Backwards compatibility is always nice, but having the component work for the latest-ish version is more important imo. I'd be fine with depending on React 15.5.x.
3) Do you think having the same approach as for _ isValidDate_ would be ok? I also think this could be improved.

bjbrewster:
Hi,

Thanks for the feedback.  I had a few TODOs I wanted to finish before submitting again.  Thanks for the feedback.  I'll aim to push what I've done by the end of next week.

I don't think something like isValidDate would work for time constraints.  Calling isValidTime for every minute of the day, like its doing for each day in a year to check if year is valid is silly.  Maybe allow the user to set the time to whatever they like, and then call isValidTime for the user's callback to return true/false with an optional message?  If we had array of allowed min/max ranges then we could more easily restrict input and show allowed values.

Cheers,
Bren

PR #448, opened 2017-10-17
Rebase with upstream

lopis:
The upstream introduces a bunch of fixes including compatibility with React 16.


lopis:
Ignore. Openned in the wrong repo.

PR #455, opened 2017-11-01
Pass event to onFocus event handler

simeg:


simeg:
If #458 gets merged then we can merge this one. If not then I'll have to look into getting this change in some other way.

PR #458, opened 2017-11-04
React 16

simeg:
**I'd like some feedback on this.** This PR changes one significant thing: we're saying the component needs React with version >=16. I am not sure what this PR changes for all consumers of this component, hence my call for help.

_Background taken from a commit message_
Some background: I wanted to make a change and I could not get Travis to
run the tests without failing. I did some research and it turns out the
test util package enzyme made some major changes which broke our stuff.
At least that's my theory. I did some testing and the only solution I
found was supporting React from 16.0.0 and above.

Before we supported React from 0.13 and above, and this commit breaks
that backward compability. This is unfortunate but I don't know any
other way of resolving this. I had to exclude some tests because of the
changes for enzyme but I'll take a look at them in the future.

I spent two nights debugging this and I ran out of ideas. I'm not loving this solution but as I said I don't know what else to do. If someone could explain the impact of this I'd be thankful.

@layneanderson @arqex or anyone else.


layneanderson:
Hi @simeg ,

I really don't want to see this become dependant on v16 of React.  I feel it would freeze out too many people.

I'm finally going to get some time this week to look at react-datetime again (sorry I've been overloaded the last couple of months).  Thanks for all the work you've done this past weekend.  I'll try respond on this later this week.

Cheers,

Layne

bjbrewster:
Hi,

I don't think we need to go straight to React 16.  You can try updating enzyme to ^2.9.1 that has React peer dependency of `"0.13.x || 0.14.x || ^15.0.0-0 || 15.x"`.  

If we need the latest Enzyme, there is an Enzyme adapter `enzyme-adapter-react-15.4` for React versions `"15.0.0-0 - 15.4.x"` separate to the adapter `enzyme-adapter-react-15` for React version `"^15.5.0"`.

Also, the peerDependency React version can be different to the devDependency version that is used for running the tests.  Just because it's using a newer React to run the tests doesn't mean people are forced to use the same version.

Cheers,
Bren

simeg:
@bjbrewster It's true that dev dependency doesn't need to be same as peer dependency, I didn't think of that. I think I tried that version of enzyme but could be worth testing again, feel free to use my branch for testing thing if someone beats me to it. Why I didn't want to use an adapter was that I didn't know what version was gonna get resolved by npm, but if it can differ from the peer dependency then we could pin it to a major version. Also, [this](https://github.com/airbnb/enzyme/issues/1325) could maybe be of help.

[Here's a job that fails on Travis](https://travis-ci.org/YouCanBookMe/react-datetime/jobs/297688985). 

Thanks for your input, both of you.

simeg:
See #463.

PR #463, opened 2017-11-11
Let's make the build green

simeg:


simeg:
@layneanderson @bjbrewster

I listened to your opinions and advice and this is where I landed. I think this is OK for now, using react ^15.5 during testing. I had to exclude two tests and change one slightly (it's behaving differently depending on the node version, for some reason).

Thoughts?

layneanderson:
LGTM

I ran the tests on my machine and got the same results that you did.  (Although I still have to switch my timezone to Stockholm).  I wanted to do a deeper dive but my schedule got busier than I expected this week.

I'm good with React 15.5.  I would think that most are at that level or higher.  There might be a few hold outs at 14 but I think it's time to move forward.

Thanks for your great work.

bjbrewster:
It looks good to me.  Awesome work. 👍

simeg:
Great, thanks for reviewing. I'll merge this + all my others PRs when I get home later today.

Also, we should do something about the timezone thing. You should not have to change your timezone @layneanderson.

PR #469, opened 2017-11-14
Check open state in componentWillReceiveProps

fetimo:
### Description
Check to see if `this.props.open` has changed between current props and next props.

### Motivation and Context
* Updating `open` via state doesn't trigger the datepicker dialog to display, it was only checked on component construction.

### Checklist
I'm afraid I haven't added a test for this behaviour as I haven't got any Jest experience. If you're able to advise it would be greatly appreciated.


simeg:
Hi @fetimo. Sorry for the delay. This has been fixed in an upstream PR, I believe. Try out the latest version and feel free to @ me if it has not been fixed.

PR #473, opened 2017-11-27
Allow specifying viewDate as a prop, fixes #468

### LINKED ISSUE #468: Open calendar to specific date (without *setting* the date)
### I'm Submitting a ...

```
[ ] Bug report
[x] Feature request
[ ] Support request
```

Here's my use case: I have two datepickers where people can pick the start and finish dates for a trip. If we're currently in November and people select "June 2nd" as the start date for their trip, I would like the second datepicker to automatically open in June, and not in November like it does by default. 

Is there a way to achieve this without having to actually set the `value` prop for the second datepicker?

vishy1618:
### Description
Optionally specify the month that is viewed when opening the date picker without having an explicit value set.
The default behavior of `viewDate` is retained as well.

### Motivation and Context
* This allows users to open the calendar at a specific month without setting the value. This is useful when we want the user to start from/choose a specific month.
* This fixes https://github.com/YouCanBookMe/react-datetime/issues/468 

### Checklist

```
```


vishy1618:
Hi, any updates on this? Would you require anything in addition to make the PR more complete?

simeg:
Hey @vishy1618. Sorry about the delay! Thanks you for your contribution. Everything looks good so I will merge this.

keithpickering:
Hey @vishy1618 could you let me know if I'm doing something wrong? `viewDate` doesn't work at all for me. See https://github.com/YouCanBookMe/react-datetime/issues/644

Issue #343, opened 2017-06-02
Set state from updated props


Krzyku:
### I'm Submitting a ...
```
[x] Bug report
[ ] Feature request
[ ] Support request
```

### Steps to Reproduce
Wrap DateTime component in Your custom component. Now try to toggle DateTime `open` prop.

### Expected Results
Show and hide datepicker

### Actual Results
Nothing happen

### Minimal Reproduction of the Problem
https://codepen.io/krzyk/pen/JJPxyY

### Other Information (e.g. stacktraces, related issues, suggestions how to fix)
In `DateTime.js` change line inside `componentWillReceiveProps` method from `updatedState.open = this.state.open;` to `updatedState.open = nextProps.open;`

LuciRebecca:
I ran into this today as well and the above resolution from @Krzyku worked for me

MatthiasDebaere:
We also ran into this today. Can we get this merged, please?

fernandops26:
since that version?

NullVoxPopuli:
this is the technique I use for opening the picker:
```js
import React, { Component } from 'react';
import PropTypes from 'prop-types';
import ReactDateTime from 'react-datetime';
import { InputGroup } from 'react-bootstrap';

const InputAddon = InputGroup.Addon;

export default class DatePicker extends Component {
  constructor(props) {
    super(props);

    this.pickerRef = null;
  }

  render() {
    const open = () => this.pickerRef && this.pickerRef.openCalendar();

    return (
      <InputGroup>
        <ReactDateTime
          { ...this.props }
          ref={ref => { this.pickerRef = ref; }}
        />

        <InputAddon onClick={open}>
          <i className='fa fa-calendar' aria-hidden='true'></i>
        </InputAddon>
      </InputGroup>
    );
  }
}

```

SachaG:
Still doesn't seem to be working for me with the latest version.

eyanez1230:
Does not work for me either with the latest version

simeg:
@Krzyku @LuciRebecca @MatthiasDebaere @fernandops26 @NullVoxPopuli @SachaG @eyanez1230 

Please try version 2.11.1 and let me know if it worked for you. Thanks.

eyanez1230:
@simeg Confirming that it is now fixed! Thank you very much.

SachaG:
[...]

PR #476, opened 2017-12-01
Version 2.11.1

simeg:


simeg:
Should fix #343 and many more.

PR #483, opened 2017-12-06
Correct default value of `strictParsing` in README

bradchristensen:
### Description
This commit updates the README to reflect the current behaviour of the `strictParsing` prop, which is set to `true` by default.

See: https://github.com/YouCanBookMe/react-datetime/blob/d0b63bcd5330053a64331014be9e118c66205e80/DateTime.js#L49

### Motivation and Context
The documentation for this prop is currently incorrect :smile:

### Checklist
```
  [ ] I have updated the TypeScript 1.8 type definitions accordingly (not required)
  [ ] I have updated the TypeScript 2.0+ type definitions accordingly (not required)
```


simeg:
Oops. Didn't mean to do that.

@bradchristensen Nice catch, thanks for this!

PR #496, opened 2018-01-23
Use static property defaultProps instead of getDefaultProps.

MatteoMilesi:
I removed the getDefaultProps function from Datetime declaration and defined the Datetime.defaultProps property after it.

From React 15.6+, using getDefaultProps generates console warnings like this:
`Warning: getDefaultProps is only used on classic React.createClass definitions. Use a static property named 'defaultProps' instead.`


simeg:
Thanks for this, the code looks good. Even though the CI is not green I know for a fact that this test was flaky and I've fixed it since. So I will merge this anyway.

Issue #478, opened 2017-12-04
Allow next and prev arrows render


MarciDavid:
Hi

You have a really good component here.
The implementation and adjustments were very easy and the documentation is very clear.

We want to change the style a bit, is there a change to allow renderArrow function?

Thanks a lot

simeg:
Hi @MarciDavid.

There is no prop to change the arrows, however you can always override the CSS to make it look the way you want it to. Feel free to post a PR for it, it shouldn't be too complicated to write.

PR #507, opened 2018-02-07
Allow arrowRender

davidmarciano:
Can be handy to create a function which will render next and prev arrows with customize components.

I tried to change the next and prev arrows styling to another Icon font or even another HTML tag without success.
Those function should allow to get react component to render them as arrows presentation components.
More details in: https://github.com/YouCanBookMe/react-datetime/issues/478


simeg:
This PR has +16 −36,313. Your commit is basically deleting every file in this repo. Also the commits says `Deploy Demo Pages`, it doesn't sound like what you want to do with this PR.

davidmarciano:
Sorry about that.
That's my first time I'm asking for PR, I could be more noobie than that...

Anyway, I will fork the repository, make the changes and then ask for PR.
Hope it's a better way to do that.

PR #508, opened 2018-02-07
Add onSubtractTime and onAddTime hooks

DaanDD:
### Description
Implement hooks for onSubtractTime and onAddTime. 

### Motivation and Context
This could be used to dynamically load selectable dates for each month.

### Checklist
```
```

I have ran build but it seems to have updated more than only my changes in the dist folder.

### DIFF
--- a/DateTime.d.ts
+++ b/DateTime.d.ts
@@ -94,15 +94,15 @@ declare namespace ReactDatetimeClass {
          */
         onViewModeChange?: (viewMode: string) => void;
         /*
-         Callback trigger when the user navigates to the previous month, year or group of years.
+         Callback trigger when the user navigates to the previous month, year or decade.
          The callback receives the amount and type ('month', 'year') as parameters.
          */
-        onSubtractTime?: (amount: number, type: string) => void;
+        onNavigateBack?: (amount: number, type: string) => void;
         /*
-         Callback trigger when the user navigates to the next month, year or group of years.
+         Callback trigger when the user navigates to the next month, year or decade.
          The callback receives the amount and type ('month', 'year') as parameters.
          */
-        onAddTime?: (amount: number, type: string) => void;
+        onNavigateForward?: (amount: number, type: string) => void;
         /*
          The default view to display when the picker is shown. ('years', 'months', 'days', 'time')
          */
--- a/DateTime.d.ts
+++ b/DateTime.d.ts
@@ -94,6 +94,16 @@ declare namespace ReactDatetimeClass {
          */
         onViewModeChange?: (viewMode: string) => void;
         /*
+         Callback trigger when the user navigates to the previous month, year or group of years.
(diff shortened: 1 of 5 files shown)


simeg:
Hi @DaanDD and thanks for contributing! Your code is changing a critical part of the component but you didn't write any tests. I would like to see some tests that verifies that it's working as you want it to work.

Also, no need to update the dist files. We do that when we release a new version, please remove them from your PR. Thanks!

DaanDD:
Hello, thanks for the feedback! I have removed the dist files and added some tests.

DateTime.d.ts
```diff
@@ -93,6 +93,16 @@ declare namespace ReactDatetimeClass {
...
+        /*
+         Callback trigger when the user navigates to the previous month, year or group of years.
+         The callback receives the amount and type ('month', 'year') as parameters.
+         */
+        onSubtractTime?: (amount: number, type: string) => void;
+        /*
+         Callback trigger when the user navigates to the next month, year or group of years.
+         The callback receives the amount and type ('month', 'year') as parameters.
+         */
+        onAddTime?: (amount: number, type: string) => void;
```
simeg: These name has `time` in them, but they are triggered for "months, year or group of years". It might be confusing for the user since the component has a time mode.

Naming is hard for these triggers, `onSubtractMonthsOrYears` is more correct but it's long.

What are your thoughts?

test/tests.spec.js
```diff
@@ -1062,6 +1062,64 @@ describe('Datetime', () => {
 
 	});
 
+	describe('onSubtractTime', () => {
```
simeg: You are testing `onAddTime` here and `onSubtractTime` in the next `describe()`, a mixup :)

DateTime.d.ts
DaanDD: `onSubtractMonthsOrYears` seems a bit long, maybe something like `navigatePrevious`

simeg: I like that better, because navigating time doesn't really make sense. Let's go with that, what do you say?

DaanDD:
@simeg I have changed the hook names. `onSubtractTime` and `onAddTime` are now called `onNavigateBack` and `onNavigateForward`

simeg:
Looks good 👍 Thanks for contributing!

laddhadhiraj:
I am trying to call "onNavigateBack" and "onNavigateForward" both are not working.  "onViewModeChange" is working well.

`<Datetime input={false} timeFormat={false} onChange={this.updateDate.bind(this)} renderDay={ this.renderDay } onViewModeChange={this.dateViewMode.bind(this)} onNavigateBack={this.dateNavigate.bind(this)} onNavigateForward={this.dateNavigate.bind(this)}/>`

laddhadhiraj:
It worked for me If I install this package like ` npm install  --save  YouCanBookMe/react-datetime`

Issue #510, opened 2018-02-08
calendar.openCalendar() doesn't seem to work


SachaG:
### I'm Submitting a ...

```
[x ] Bug report
[ ] Feature request
[ ] Support request
```

I'm trying to manually trigger a calendar to open. I've added a `ref` to the component:

```
<DateTimePicker
  ref={(ref) => { this.toCalendar = ref; }}
/>
```

And I would then expect to be able to do `this.toCalendar.openCalendar()` to open the calendar; but that doesn't do anything. The weird part is that I can log out `this.toCalendar` to my console, and if I call `openCalendar()` from there it *does* work. 

This might very well be a problem with my code and not with the package, so if anybody has suggestions on what I could try to debug the issue that would be very helpful.

simeg:
Sounds weird. You can pass `open: true` for it to open, I believe.

SachaG:
Yes, but then I run into the other issue (https://github.com/YouCanBookMe/react-datetime/issues/509).

simeg:
Ouch. Would you be able to provide a CodePen example of this behaviour? [This one](https://codepen.io/simeg/pen/mEmQmP) has the latest react-datetime code loaded.

SachaG:
A minimal example works: https://codepen.io/SachaG/pen/qxRQbb

So it's probably something else in my code… I'll keep looking :)

SachaG:
OK, I've narrowed it down: apparently `openCalendar()` stops working if you do a `setState` in the same function: https://codepen.io/SachaG/pen/zRZWZg?editors=0110

simeg:
Wow, great work narrowing down the issue. Would you mind submitting a PR with a fix? (and maybe a test or two, if possible)

SachaG:
I'm not very familiar with the code yet so until I have time to submit a proper PR I made one to add some documentation about `openCalendar`: #513

SachaG:
Closing this because we decided using `ref` is an anti-pattern. If you do need to use it, here are some additional details (use at your own risk!):

---

# Manually Opening the Calendar

In some cases, you'll want to manually get the calendar to open. While you can do so by setting `open` to `true`, another approach is to call the calendar's `openCalendar` method. For that, you'll need to make the calendar available via its `ref` property: 

```js
class Calendar extends React.Component {
  constructor() {
    super();
  }

  render() {
    return (
      <div>
        <button onClick={this.datetime.openCalendar}>Open</button>
        <Datetime
[...]

PR #513, opened 2018-02-11
Add instructions for openCalendar

SachaG:
Document how to use openCalendar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,39 @@ class MyDTPicker extends React.Component<MyDTPickerProps, MyDTPickerState> {
 }
 ```
 
+## Manually Opening the Calendar
+
+In some cases, you'll want to manually get the calendar to open. While you can do so by setting `open` to `true`, another approach is to call the calendar's `openCalendar` method. For that, you'll need to make the calendar available via its `ref` property: 
+
+```js
+class Calendar extends React.Component {
+  constructor() {
+    super();
+  }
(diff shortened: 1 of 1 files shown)


simeg:
I'm not sure I want to "encourage" using `ref` with the component. If you are using `ref` with a React component it's usually something wrong, so let's try to fix what's wrong instead of encouraging bad practice.

What do you say? @SachaG

SachaG:
I don't disagree, but then what would the ideal API look like?

SachaG:
I guess we could just rely on `open` instead. My main reason for wanting to use `openCalendar` was to differentiate between "I want to trigger this calendar's opening" and "I want this calendar to remain in an open state", which I thought was more what `open` was for. But maybe that's not really a meaningful distinction.

simeg:
I think it is a meaningful distinction, but can't this be done today _with_ the `open` prop? If you consider it to be a state it makes sense to use it for both, then you could just write your own `openCalendar`:

```js
const openCalendar = bool => bool;
<Datetime
  open={openCalendar}
/>
```

SachaG:
True. Is #510 worth fixing then? Or should people just not use `openCalendar`?

simeg:
If your issue is solved, then feel free to close it. I'd not advise people to use that function using refs.

SachaG:
OK.

PR #514, opened 2018-02-11
Dynamic viewDate

### LINKED ISSUE #468: Open calendar to specific date (without *setting* the date)

SachaG:
Make `viewDate` update when prop changes. Fixes #468.

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -183,6 +183,10 @@ var Datetime = createClass({
 				}
 			}
 		}
+
+		if ( nextProps.viewDate !== this.props.viewDate ) {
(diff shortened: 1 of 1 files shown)


simeg:
Great @SachaG, thank you 🙏

PR #515, opened 2018-02-11
React 16 Update (WIP)

SachaG:
While working on #514 I had to update to React 16 to make the code build properly. As far as I can tell the package itself works, but the tests don't run yet. So this is just a WIP in case somebody else wants to build on top of that. 

Feel free to close this if it's not helpful :)


arqex:

PR #520, opened 2018-02-13
Only call hooks when setState has finished updating the internal state

DaanDD:
### Description
Before this, when a hook like `onViewModeChange`, `onNavigateBack` or `onNavigateForward` was called, the internal state would not yet be updated. This meant that when I tried to access the current `viewDate` through the internal state after one of these hooks was called, the setState would not have been updated yet, because `setState` is asynchronous. This PR only calls these hooks when `setState` has completed.

### Checklist
```
```

The tests had to be refactored a bit. I had to implement an async/await pattern that waits for the `setState` to complete before moving on to the next test. For some reason all tests within the hooks are ignored without the async/await pattern. This also meant I had to upgrade the eslint and the eslint parser for the tests to ES2017. This also meant I had to fix some issues in the code itself.

### DIFF
--- a/test/.eslintrc.js
+++ b/test/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
 		"parserOptions": {
-				"ecmaVersion": 6,
+				"ecmaVersion": 2017,
 				"sourceType": "module",
 				"ecmaFeatures": {
 						"jsx": true
--- a/DateTime.js
+++ b/DateTime.js
@@ -235,8 +235,10 @@ var Datetime = createClass({
 	showView: function( view ) {
 		var me = this;
 		return function() {
-			me.state.currentView !== view && me.props.onViewModeChange( view );
-			me.setState({ currentView: view });
+			var previousView = me.state.currentView;
+			me.setState({ currentView: view }, function() {
+				previousView !== view && me.props.onViewModeChange( view );
+			});
 		};
 	},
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -51,10 +51,11 @@
     "babel-loader": "^6.2.1",
     "babel-plugin-transform-remove-strict-mode": "0.0.2",
     "babel-preset-es2015": "^6.22.0",
+    "babel-preset-2017": "0.0.0",
     "babel-preset-react": "^6.22.0",
     "enzyme": "^3.0.0",
     "enzyme-adapter-react-15": "^1.0.5",
-    "eslint": "^3.1.0",
+    "eslint": "^4.17.0",
```
simeg: There seems to be a lot of changes completely unrelated to the commit message in this commit. It would be nice to have them separated, like why is this dependency updated? And why is `babel-preset-2017` introduced?

DaanDD: It seems like I committed a bit too much. I thought the es2017 preset for `eslint` could only work in newer versions, but turns out it didn't. The `babel-preset-2017` also didn't belong there. I have cleaned it up accordingly.

test/.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
 		"parserOptions": {
-				"ecmaVersion": 6,
+				"ecmaVersion": 2017,
```
simeg: Why is this needed?

DaanDD: Because I needed to use async/await in the tests to wait for the `setState` callback to complete.

arqex:

PR #523, opened 2018-02-21
Close on selection of month instead of year when dateFormat is set to display quarter

MilosRasic:
### Description

Altered getUpdateOn() method to return viewModes.MONTHS when Q token is used in dateFormat. In case of dateFormats like YYYY-[Q]Q, it would detect Y token and return 'years' instead, but month selection is required to specify a quarter of a year.

Added a test case to cover this but also one to cover behaviour when w token is used (week of year), which is handled correctly even though it's not explicitly matched because getUpdateOn() returns viewModes.DAY by default.

### Motivation and Context

My team ran into this problem while using react-datetime with closeOnSelect=true to control data for rendering of charts in which date grain can also be selected in another component. Changing date grain updates dateFormat prop on datepickers. It works fine for all formats except YYYY-[Q]Q where it closes when year is selected, making it impossible for user to get to months view.

Also added a test case to check week of the year format (for example gggg-[W]ww), which is also something we use. This is currently working correctly because all token detection fail and the method returns 'days' by default. I feel however that it would be easy to break with future changes since it relies on the default return, hence a new test case.

### Checklist

```
[ x ] I have not included any built dist files (us maintainers do that prior to a new release)
```

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -105,7 +105,7 @@ var Datetime = createClass({
 	getUpdateOn: function( formats ) {
 		if ( formats.date.match(/[lLD]/) ) {
 			return viewModes.DAYS;
-		} else if ( formats.date.indexOf('M') !== -1 ) {
+		} else if ( formats.date.match(/[MQ]/) ) {
 			return viewModes.MONTHS;
 		} else if ( formats.date.indexOf('Y') !== -1 ) {
 			return viewModes.YEARS;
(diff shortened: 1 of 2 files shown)


arqex:
Hey  @MilosRasic 

Sorry for the delay. Now in v3 there is a `updateOnView` property that does exactly what you want, no need to mess with the date format:

```jsx
import React from "react";
import "react-datetime/css/react-datetime.css";
import Datetime from "react-datetime";

export default function App() {
  return (
    <Datetime updateOnView="months" timeFormat={false} closeOnSelect={true} />
  );
}
```

See it working: https://codesandbox.io/s/optimistic-bartik-mnxt2

PR #529, opened 2018-03-01
support passing attributes to DOM nodes for prev/switch/next buttons and calendar tiles

MilosRasic:
Added more props similar to inputProps to support passing DOM attributes to nodes corresponding to prev, switch, next buttons and calendar tiles.

### Description
Added new props corresponding to following DOM nodes:
prevButtonProps => .rdtPrev
switchButtonProps => .rdtSwitch
nextButtonProps => .rdtNext
tileProps => td

Each prop can be an object or a function returning an object. Each key in the object will be passed as a prop to the corresponding DOM node.

Functions for buttons receive the current view mode as an argument.

tileProp function receives current view mode, moment object corresponding to the tile and an isDisabled flag as arguments.

Static object prop or the function prop return is merged with default hardcoded props. For this purpose, the function in mergeProps.js is introduced.

As side-effect, viewModes object is extracted into a separate file so that it can be imported in views.

### Motivation and Context
My team ran into a need to customize attributes on some of the DOM nodes rendered by react-datepicker. Specifically, we want to put data attributes on the nodes covered by this PR as markers for end to end tests.

Depending on needs, the support for custom attributes could be expanded to cover more DOM nodes. We could also pass more arguments to function props.

### Checklist
```
[ x ] I have not included any built dist files (us maintainers do that prior to a new release)
[ x ] My changes required the documentation to be updated
  [ x ] I have updated the documentation accordingly
  [ x ] I have updated the TypeScript 1.8 type definitions accordingly
  [ x ] I have updated the TypeScript 2.0+ type definitions accordingly
```


arqex:
Sorry for getting late into this PR but thanks!. Many changes happened since it was created and now it's possible to completelly customize the views using the `renderView` prop.

PR #531, opened 2018-03-06
Changed compare operator from === to ==

borovik96:
image

### Description

`undefined !== null`, so 
```javascript
var value = e.target === null ? e : e.target.value
``` 
throws an error in case e.target is undefined


arqex:

PR #539, opened 2018-03-19
Add renderInput property to definition

Ritorna:
In the current version the renderInput property is missing in the definitions. 
For adding custom Rendering (for example Material Ui Input)  we need you to provide that property in the typings.
Works fine with edited typings, see: 
image


arqex:

PR #536, opened 2018-03-15
Added renderInput definition to typings

Kurounin:
### Description
Added missing renderInput property to typings

### Motivation and Context
Solves typescript 2.0+ errors when trying to use renderInput prop.

### Checklist
```
[x] I have not included any built dist files (us maintainers do that prior to a new release)
[ ] I have added tests covering my changes
[x] All new and existing tests pass
[ ] My changes required the documentation to be updated
  [ ] I have updated the documentation accordingly
  [ ] I have updated the TypeScript 1.8 type definitions accordingly
  [x] I have updated the TypeScript 2.0+ type definitions accordingly
```

PR #533, opened 2018-03-12
Adding renderInput to typescript definition

ElysiumTan:
Adding renderInput to typescript definition

### Description
See above, simple as that.  Includes props, openCalendar, closeCalendar, and description.

### Motivation and Context
Missed definition -> see #482

### Checklist

```
[ ] I have not included any built dist files (us maintainers do that prior to a new release)
[ ] I have added tests covering my changes
[ ] All new and existing tests pass - looks like some tests are failing due to failure to update a snapshot, if this is a huge issue, can just submit it with a new snapshot, but feel like that'd be something covered by a different PR?
[ ] My changes required the documentation to be updated
  [ ] I have updated the documentation accordingly
  [ ] I have updated the TypeScript 1.8 type definitions accordingly
  [ ] I have updated the TypeScript 2.0+ type definitions accordingly
```

PR #541, opened 2018-03-24
Add renderInput to TypeScript definition

paul-roman:
### Description
I added the following definition of the `renderInput` prop to the TypeScript definition file:
```typescript
renderInput?: (props: any, openCalendar: () => void, closeCalendar: () => void) => void;
```

### Motivation and Context
* This pull request is made because I used to work with this project for a TypeScript-based application and I see that the definition for the `renderInput` prop was missing, causing my project to fail transpiling.
* I think it should be merged to help other people like me, who are using the TypeScript version of the component and that want to use this prop.

### Checklist
```
```

### DIFF
--- a/DateTime.d.ts
+++ b/DateTime.d.ts
@@ -132,6 +132,13 @@ declare namespace ReactDatetimeClass {
          */
         renderDay?: (props: any, currentDate: any, selectedDate: any) => JSX.Element;
         /*
+         Replace the rendering of the input element. The function has the following arguments:
+         the default calculated props for the input, openCalendar (a function which opens the calendar)
+         and closeCalendar (a function which closes the calendar).
+         Must return a React component or null. See input customization
+         */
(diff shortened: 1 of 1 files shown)


Ritorna:
Also see #539 #536 #533.
@simeg, @YouCanBookMe can someone of you please review those?

seansfkelley:
Since the maintainers seem to be AFK, you can do this where you need the type like so:

```
const TypedDatetime = Datetime as React.ComponentType<DatetimepickerProps & {
  renderInput?: (props: any, openCalendar: () => void, closeCalendar: () => void) => void;
}>;
```

I tried to define the behavior using module augmentation, but the `export =` syntax and `namespace` make it pretty messy and I ended up with the (kinda lame) solution above.

kevinrambaud:
@arqex any chance to take a look?

arqex:

PR #554, opened 2018-05-18
Update DateTime.d.ts

vasilevich:
Just made a tiny update to index.d.ts added "name" definition which works, but was just not defined in the .ts file , it is useful on certain cases where I need to detect the exact element type in the ref I defined earlier, thus allowing me to easily parse the input by its type.


ESWAT:
Can we get this merged? Simple solution to address TS build warnings that don’t need to happen.

arqex:

PR #575, opened 2018-07-25
Create disableEventPropagation prop to avoid external event side-effects

troycroz:
### Description
We wrap react-datetime in a Semantic UI modal, and sometimes (particularly the TimeView clicks) react-datetime click propagation would cause our modals to behave irregularly.

This should allow users of react-datetime to bound events to this component, and avoid them creating external side-effects in the client application.

### Motivation and Context
Explained above.

### Checklist
```
```

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -15,6 +15,13 @@ var viewModes = Object.freeze({
 	TIME: 'time',
 });
 
+var stopEventPropagation = function ( e ) { e.nativeEvent.stopImmediatePropagation(); };
+var eventPropagationHandlers = Object.freeze({
+	onClick: stopEventPropagation,
+	onKeyUp: stopEventPropagation,
+	onTouchEnd: stopEventPropagation
+});
+
 var TYPES = PropTypes;
 var Datetime = createClass({
 	displayName: 'DateTime',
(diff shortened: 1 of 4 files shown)


arqex:
Hey @troycroz 

Thanks for the PR and for your efforts in doing things right! I'll have a look at it.

troycroz:
Thanks @arqex.

My only concern was how best to test this. Ideally I should be able to assert that all of the existing tests still work in this disableEventPropagation mode, so that is the tact that I took. However, that required me stub out out the nativeEvent's stopImmediatePropagation when the test clicks are simulated, which is a bit clunky.

In a perfect world the tests would run once with this prop and once without it, but I didn't want to double the size of the test suite.

anushkadoyan:
Was this merged?

troycroz:
For what it's worth, Contactually has been running a fork of this branch stably in production for several months now.

MightyDespair:
I'm using datetime in a semantic ui modal and experiencing the same problem with the time view. What's happening with this PR?

arqex:
Sorry for getting late into this PR but thanks! Best way of prevent event bubbling is not by updating react-datetime (some other people would like to listen to that event). If it's needed you can wrap the compoent into a `div` and prevent the event propagation there:

```jsx
<div onClick={ e => e.stopPropagation() }>
  <Datetime />
</div>
```

PR #594, opened 2018-10-18
Add position override

joshleblanc:
Add a prop the override the calendar's position style. It defaults to `absolute`, which isn't always what you might want. 

### Description
Add a prop the override the calendar's position style. It defaults to `absolute`, which isn't always what you might want.

### Motivation and Context
We have a project that's styled completely with jss. There was no way for me to change the positioning to relative in this manner.

### Checklist

```
```


arqex:
Sorry to get late to this PR. Styles are completely overridable by using CSS. If you can't use CSS, we are not going to add props for individual style properties, we better need to think a way of passing any style to the internal layers. Closing.

PR #600, opened 2018-11-10
Allow strictParsing when setting displayTimeZone

summera:
### Description
Currently, when setting `displayTimeZone` strict parsing will not work because `props.strictParsing` is not being passed to `moment.tz`. `moment.tz` is able to accept a boolean to enable/disable strict parsing [according to the docs](https://momentjs.com/timezone/docs/#/using-timezones/parsing-in-zone/).

### Motivation and Context
See above

### DIFF
--- a/DateTime.js
+++ b/DateTime.js
@@ -396,7 +396,7 @@ var Datetime = createClass({
 		if (props.utc) {
 			m = moment.utc(date, format, props.strictParsing);
 		} else if (props.displayTimeZone) {
-			m = moment.tz(date, format, props.displayTimeZone);
+			m = moment.tz(date, format, props.strictParsing, props.displayTimeZone);
 		} else {
 			m = moment(date, format, props.strictParsing);
 		}


LincolnPuzey:
This bug is also affecting me. Can this be merged?

mk0003:
I tried and my problem with defaultTimeZone prop fixed. It would be good to get this merged.

arqex:
Sorry to get late to this PR, but thanks for it! Now the code is updated and the issue has already been fixed in the last version. Closing it.

mk0003:
@arqex I tested the timezone strictParsing bu it doesn't work. I checked the source code but it doesn't look like this code merged.
This is the code line i checked:
v3 branch:
https://github.com/arqex/react-datetime/blob/d619062a6546b191a66db6d1674ed3d6fe2ba844/src/datetime/DateTime.js#L426-L432
master branch:
https://github.com/arqex/react-datetime/blob/d6e72d27042c0f9201b02348c0e1d76b828f65a1/src/datetime/DateTime.js#L427-L433

PR #568, opened 2018-07-04
Add ability to customize first day of week

summera:
### Description
This allows you to customize the first day of the week in `react-datetime` by passing in the first day of the week as a string.

```js
<Datetime firstDayOfWeek="sunday" />
```

### Motivation and Context
This is independent of locale. So if a users locale says the first day of the week is `Sunday`, but you would like to force it to be `Monday` in the datetime picker, this allows you to do so. In addition if you want to customize based on some setting, this allows you to do that as well. We are using something like this in our app so that the user can set their first day of the week in their profile settings and this is then reflected in a calendar view and also in the datetime pickers throughout the app.

### Checklist

```
[x] I have not included any built dist files (us maintainers do that prior to a new release)
[ ] I have added tests covering my changes
[ ] All new and existing tests pass
[x] My changes required the documentation to be updated
  [x] I have updated the documentation accordingly
  [x] I have updated the TypeScript 1.8 type definitions accordingly
  [x] I have updated the TypeScript 2.0+ type definitions accordingly
```

Note: I have **not** added tests yet but would be more than happy to do so if this is something you all are willing to accept 👍 .

PR #601, opened 2018-11-10
Add ability to customize first day of week

summera:
### Description
This allows you to customize the first day of the week in `react-datetime` by passing in the first day of the week as a string. 

```js
<Datetime firstDayOfWeek="sunday" />
```

### Motivation and Context
This is independent of locale. So if a users locale says the first day of the week is `Sunday`, but you would like to force it to be `Monday` in the datetime picker, this allows you to do so. In addition if you want to customize based on some setting, this allows you to do that as well. We are using something like this in our app so that the user can set their first day of the week in their profile settings and this is then reflected in a calendar view and also in the datetime pickers throughout the app.

### Checklist

```
```

@arqex this is the same as https://github.com/YouCanBookMe/react-datetime/pull/568 but with tests.

### DIFF
--- a/DateTime.d.ts
+++ b/DateTime.d.ts
@@ -170,6 +170,10 @@ declare namespace ReactDatetimeClass {
          close it.
          */
         disableOnClickOutside?: boolean;
+        /*
+         Customize the first day of the week in the date picker (`monday`, `tuesday`, `wednesday`, `thursday`, `friday`, `saturday`, `sunday`).
+         */
+        firstDayOfWeek?: string;
     }
 
     export interface DatetimepickerState {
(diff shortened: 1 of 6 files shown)


summera:
@arqex checking back on this and on https://github.com/YouCanBookMe/react-datetime/pull/600

summera:
@arqex any chance we can get these merged in?

LambertArno:
Is there any progress for this issue? I see it passed all the checks. Could it possibly be merged?

summera:
@LambertArno I've been trying to get this in for almost a year now. Unfortunately, no response from their team. We've had to use a fork of the library.

arqex:
Sorry to get late to this PR, and now the PR has gotten a bit obsolete.

The official way of getting updating the first day of the week is using a locale where the first day of the week is Monday. Right now if we want to customize the first day without the local, we would need to use the `renderView` prop to override the day view, and display the the days the way we need.

I'm going to close this PR for the time being, but I'm open to re-think adding the prop if there is more requests for it.

PR #603, opened 2018-11-20
FEATURE: add a snap flag to enable snapping to plausible steps on inc & dec in TimeView

JamesAlias:
### Description

I'm referencing the original issue from another repo here: neos/neos-ui#2247 to see where the idea originated.

I added a feature behind a flag called `snap` which _when enabled_ slightly alters the way the `increase` and `decrease` methods of `TimeView` work.
The flag is optional and completely backwards compatible so there are no breaking changes.

### Motivation and Context

When the user uses the `today` button he sets the time to _now_ which, for example, could be `00:12`. 
With `step` for `minutes` set to `5` the `increase` or `decrease` buttons on minutes it will step to `17` or `7` respectively.
With `snap` enabled it will _snap_ to the next _plausible_ value depending on `step`, e.g. `15` and `10` respectively.

### Checklist

```
```

### DIFF
--- a/DateTime.d.ts
+++ b/DateTime.d.ts
@@ -16,9 +16,9 @@ declare namespace ReactDatetimeClass {
     export type ViewMode = "years" | "months" | "days" | "time";
 
     export interface TimeConstraint {
-        min: number;
-        max: number;
-        step: number;
+        min?: number;
+        max?: number;
+        step?: number;
     }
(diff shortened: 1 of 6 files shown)


jonnitto:
@YouCanBookMe is there a chance that you will merge this and make a release?

PR #616, opened 2019-01-24
Add feature to render custom time arrows

danlupascu:
### Description
This pull request adds an extra prop to the Datetime component that allows to render custom time arrows for the time view.

### Motivation and Context
There is no way to change the arrows for the time view. This
feature allows to change any of the 8 possible arrows on the time view.

### Checklist
```
```


arqex:
In the new v3, it's possible to use your own components to customize the time view through the `renderView` prop. Closing here.

PR #623, opened 2019-02-23
refactored into typescript

capaj:
this is probably not going to get merged, but what the heck-if there is a slim chance let's try at least.

I've rewritten the codebase to typescript. Next on the roadmap would be to get rid of `react-onclickoutside` as a dependency.

I will try to bring in the latest changes from the upstream and then I'll mark as ready for code review.


arqex:
Hey @capaj this is one of the most fabulous PR this libraries had ever had!

I'm arriving late to it, and now it's too late to start merging it. One of the ambitions for v4 is to port the source to typescript, so I won't close this PR in case we can salvage some of this work for the new version.

In that new version we are looking for more people to collaborate, and since you already know about the internals of react-datetime, it would be a pleasure if you can add your 2 cents to the roadmap of the new version:

https://github.com/arqex/react-datetime/issues/723

Thanks so much for your effort.

PR #655, opened 2019-09-17
Check if [date] is given

termosa:
Added check for `date` to be defined.

### Description
Checked for `date` presence in method where props are processed.

### Motivation and Context
On processing props `date` was not checked for presence and the property `date.isValid` is used. This cause an error if `value` and `defaultValue` are not set.

### Checklist
```
```


arqex:
Sorry to get late to this PR. Now the code is updated and the issue has already been fixed in the last version. Closing it.

PR #656, opened 2019-09-17
Modify time-picker to display additional zero before hours

ievgenii-manzyk:
Modify time-picker to display additional zero before hours. This should apply only for numbers from 0 to 9.

### Motivation and Context:
We can format date in text field in different ways, let's see an example:
09/17/2019 **03**:28 AM
So it would be great to add additional '0' to hours in time-picker.

### Checklist
[x ] I have not included any built dist files (us maintainers do that prior to a new release)
[x ] My changes required the documentation to be updated

### DIFF
--- a/DateTime.d.ts
+++ b/DateTime.d.ts
@@ -170,6 +170,10 @@ declare namespace ReactDatetimeClass {
          close it.
          */
         disableOnClickOutside?: boolean;
+        /*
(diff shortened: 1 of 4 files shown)


arqex:
It's possible to customize the format of the date or time by using the `dateFormat` and `timeFormat` props. Closing.

PR #669, opened 2019-11-28
fix: change first-child to first-of-type in css

### LINKED ISSUE #668: Using `react-datetime` css with emotion generates a console error
### I'm Submitting a

```
[x] Bug report
[ ] Feature request
[ ] Support request
```

### Steps to Reproduce
Using `react-datetime` css with emotion generates a console error.

image

Use the following code to generate the error:

```
/** @jsx jsx */
import React from 'react';
import { jsx, css } from '@emotion/core';
import reactDateTimeStyles from 'react-datetime/css/react-datetime.css';
import DateTime from 'react-datetime';

export default class DateTimeComponent extends React.Component {
  render() {
    return (
      <div
        css={css`
          ${reactDateTimeStyles};
        `}
      >
        <DateTime value={new Date()} />
      </div>
    );
  }
}
```

### Expected Results
No error is logged

### Actual Results
Error is logged

### Minimal Reproduction of the Problem
1. [Setup a react app](https://github.com/facebook/create-react-app).
2. [Add emotion](https://emotion.sh/docs/install)
3. Add a react component using the above code.
4. Load the react app in the browser see the error in the console.

erezrokah:
Fixes #668

### DIFF
--- a/css/react-datetime.css
+++ b/css/react-datetime.css
@@ -124,10 +124,10 @@
   color: #999999;
   cursor: not-allowed;
 }
-.rdtPicker thead tr:first-child th {
+.rdtPicker thead tr:first-of-type th {
   cursor: pointer;
 }
-.rdtPicker thead tr:first-child th:hover {
+.rdtPicker thead tr:first-of-type th:hover {
   background: #eeeeee;
 }
 


arqex:
The code is updated and the issue has already been fixed in the last version. Thanks!

Issue #680, opened 2020-01-22
[v3] FileCasePathsPlugin having issues with this repo


brandon-pereira:
### I'm Submitting a ...
```
[ X] Bug report
[ ] Feature request
[ ] Support request
```

### Steps to Reproduce
This packages breaks when used with https://www.npmjs.com/package/case-sensitive-paths-webpack-plugin because the webpack externals is set to require react from React instead of react.

### Expected Results
Imports `require("React")` become "require("react")"

### Actual Results
Imports `require("React")` present 

### Minimal Reproduction of the Problem
Build the project, see the `dist/react-datetime.cjs.js` file and search for `require('React`);`

arqex:
This issue is fixed in v3. Thanks for the info!

PR #681, opened 2020-01-22
fix(webpack): fix issue with webpack building incorrect casing

brandon-pereira:
Linked to issue #680 

Note: My jest tests were failing even when clean, so I pushed with `--no-verify`. I don't have time to investigate the tests unfortunately.


arqex:
Sorry to get late to this PR. Now it's a bit updated and the issue has been already fixed. Closing it.

PR #683, opened 2020-02-11
fix: fix missing css in version 3 package

gfortaine:
### Checklist

```
[ ] I have not included any built dist files (us maintainers do that prior to a new release)
```


arqex:
Hey @gfortaine 

Thanks for the PR! I have a question, how do you import the CSS file in your project? 

In the last betas of v3 the file has been moved, but it was kind of on purpose. Depending on how people use it we can decide where it's the best place for that CSS file.

gfortaine:
@arqex We import the one that is exposed from the package. With v2, it was ```~/react-datetime/css/react-datetime.css```. We can safely assume that there is no need for a breaking change, can’t we ?

arqex:
I think you are right, there is no need of adding this breaking change.

The reason for moving the css file was that v3 now includes a testing app in it, a playground in the `/src` folder and it can't import files outside that `/src` folder. I'll fix it in the next version.

Thanks also for the updates on the readme file, they are great! This v3 is working quite right and I think docs is the last part missing before the final release.

PR #727, opened 2020-09-21
Extract ViewNavigation component

onlined:
### Description
Three files use same navigation structure, so I extracted it to a
component for DRY.

### Motivation and Context
- DRY
- Abstracting out implementation details 

### Checklist

```
```

### DIFF
--- a/src/datetime/MonthsView.js
+++ b/src/datetime/MonthsView.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import ViewNavigation from './ViewNavigation';
 
 export default class MonthsView extends React.Component {
 	render() {
@@ -6,7 +7,7 @@ export default class MonthsView extends React.Component {
 			<div className="rdtMonths">
 				<table>
 					<thead>
-						{ this.renderHeader() }
+						{ this.renderNavigation() }
 					</thead>
 				</table>
 				<table>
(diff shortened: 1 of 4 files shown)


src/datetime/MonthsView.js
```diff
@@ -22,17 +23,13 @@ export default class MonthsView extends React.Component {
 		let year = this.props.viewDate.year();
```
arqex, line 23: In some views the method to render the header is called `renderHeader` and in others `renderNavigation`. Why don't we agree on use one name for all? I'd prefer `renderNavigation`, header is too general and the `DayView` has a secondary header, so when we talk about the navigation we know what we are talking about.

I'd also call the new component `ViewNavigation` instead of just `Header`.

arqex (commented):
This PR is great and it's definitely going in the right direction! If we are going to create smaller files, we probably would like to also update the folder structure.

I was thinking about something like:
```
src/
|-- DateTime.js
|-- parts/
|   |-- DateInput.js
|   |-- ViewNavigation.js
|-- playground/
|   |-- App.js
|   |-- index.js
|-- views/
|   |-- DaysView.js
|   |-- MonthsView.js
|   |-- YearsView.js
```

With the playground we have the problem to make it work after moving their files out from the root of `/src`.
What do you think?

src/datetime/MonthsView.js
onlined: Sounds great!

onlined:
@arqex Directory structure seems good, but I think it should be organized in another PR.

src/datetime/MonthsView.js
onlined: Done!

PR #728, opened 2020-09-21
Remove unneeded parameters

onlined:
### Description
Many methods receive arguments related to object state, which they
can access. This might create confusion and inconsistency about the
source of truth. So I removed unnecessary parameters.

### Motivation and Context

- Simplifying methods
- Use component props and state as the source of truth in components 

### Checklist

```
```

### DIFF
--- a/src/datetime/DateTime.js
+++ b/src/datetime/DateTime.js
@@ -81,7 +81,7 @@ export default class Datetime extends React.Component {
 
 	constructor( props ) {
 		super( props );
-		this.state = this.getInitialState( props );
+		this.state = this.getInitialState();
 	}
 
 	render() {
@@ -89,7 +89,7 @@ export default class Datetime extends React.Component {
 			<ClickableWrapper className={ this.getClassName() } onClickOut={ this._handleClickOutside }>
 				{ this.renderInput() }
 				<div className="rdtPicker">
-					{ this.renderView( this.state.currentView, this._renderCalendar ) }
+					{ this.renderView() }
 				</div>
 			</ClickableWrapper>
 		);
@@ -121,14 +121,14 @@ export default class Datetime extends React.Component {
 		);
 	}
 
-	renderView( currentView, renderer ) {
+	renderView() {
 		if ( this.props.renderView ) {
-			return this.props.renderView( currentView, () => renderer(currentView) );
+			return this.props.renderView( this.state.currentView, () => this._renderCalendar() );
 		}
-		return renderer( this.state.currentView );
+		return this._renderCalendar();
 	}
 
-	_renderCalendar = currentView => {
+	_renderCalendar = () => {
 		const props = this.props;
 		const state = this.state;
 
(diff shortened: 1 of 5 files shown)


src/datetime/TimeView.js
```diff
@@ -246,3 +232,17 @@ export default class TimeView extends React.Component {
...
 	}
 }
+
+function pad( type, value ) {
+	const padValues = {
+		hours: 1,
+		minutes: 2,
+		seconds: 2,
+		milliseconds: 3
+	}
```
arqex, line 242: Linter say that there is a ; missing

src/datetime/YearsView.js
```diff
@@ -32,16 +31,15 @@ export default class YearsView extends React.Component {
...
 
-	renderYears( viewYear ) {
+	renderYears() {
+		const viewYear = this.getViewYear();
 		// 12 years in 3 rows for every view
 		let rows = [ [], [], [] ];
-		let selectedYear = this.props.selectedDate && this.props.selectedDate.year();
-
 		for ( let year = viewYear - 1; year < viewYear + 11; year++ ) {
 			let row = this.getRow( rows, year - viewYear );
```
arqex, line 39: We can remove `this.` here (this.getRow)

arqex (commented):
It's weird that the linter didn't catch the issues in travis, or before creating the commit.

src/datetime/TimeView.js
onlined: Added.

src/datetime/YearsView.js
onlined: Thanks, removed!

src/datetime/DateTime.js
```diff
@@ -121,14 +121,14 @@ export default class Datetime extends React.Component {
 		);
 	}
 
-	renderView( currentView, renderer ) {
+	renderView() {
 		if ( this.props.renderView ) {
-			return this.props.renderView( currentView, () => renderer(currentView) );
+			return this.props.renderView( this.state.currentView, () => this._renderCalendar() );
```
arqex, line 126: We can use `this._renderCalendar` instead of `() => this._renderCalendar()` since `_renderCalendar is already auto bound

src/datetime/DaysView.js
```diff
@@ -76,7 +75,9 @@ export default class DaysView extends React.Component {
 		));
 	}
 
-	renderDay( date, startOfMonth, endOfMonth ) {
+	renderDay( date ) {
+		const startOfMonth = this.props.viewDate.clone().startOf('month');
```
arqex, line 79: Parameter simplification is great, but in this case I think we can take `startOfMonth` and `endOfMonth` out to `renderDays`, so we clone the dates just once, instead of once for every single day and save some processing.

arqex (commented):
Today I had the time to make a good review. There is a couple of small things, when resolved we can merge :)

src/datetime/DaysView.js
onlined: What do you think about memoization? We can use a library like [`memoize-one`](https://github.com/alexreardon/memoize-one) to make these memoized methods.

arqex: I love memoize-one! I use it in all my apps. I would resist to add it as a dependency here though. At least until we clean the code and see that is really needed for some performace issue.

src/datetime/DateTime.js
onlined: Done.

src/datetime/DaysView.js
onlined: Done.

PR #732, opened 2020-09-22
Reduce impact on DOM render

GertjanVermeir:
**Description**
Multiple datePickers on one page can cause the DOM rendering to slow down.
In my current project we have tables with 20 to 100 row that can be edited and we see the DOM becomes very big.

**Motivation and Context**
- Amount of nodes in the dom before the change: 71
- Amount of nodes in the dom after the change: 2

(Done with document.querySelector(".rdt").getElementsByTagName('*').length)

**Checklist**

### DIFF
--- a/src/datetime/DateTime.js
+++ b/src/datetime/DateTime.js
@@ -59,7 +59,7 @@ export default class Datetime extends React.Component {
 		onCalendarClose: nofn,
 		onChange: nofn,
 		onNavigate: nofn,
-		onBeforeNavigate: function(next) { return next; }, 
+		onBeforeNavigate: function(next) { return next; },
 		onNavigateBack: nofn,
 		onNavigateForward: nofn,
 		dateFormat: true,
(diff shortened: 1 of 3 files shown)


arqex:
Hey @GertjanVermeir 

Thanks for the PR. This is something we wanted to tackle soon, but unfortunately we can't just return `null` when the picker is closed. There are some people that might be animating when the picker is open/closed by custom css, if we just return `null` we will be breaking the closing animation.

We would need to add kind of a timer to return null some time after the picker has been closed, but not instantly. I think the best approach would be create a new wrapper component that handled that timer. It would receive a prop `open`, if it changes from `true` to `false` it would wait a second or so (this parameter should be configurable) and then re-render to return `null`.

thomasvaneyck:
@arqex we would like to help with this issue. We are right now using a patched version internally. @GertjanVermeir clear on what to do?

PR #738, opened 2020-09-29
cleanup: remove a probably forgotten console log

istvan-derda:
### Description
removed one console.log()

### Motivation and Context
Just noticed a console output which irritated me, I guess it is a leftover debugging statement? If so I thought it's easiest to just remove it myself :)

Great work! Keep it up! :)


arqex:
Ooops! I think it was me who let that pass! Thanks for catching it!

PR #741, opened 2020-10-05
Close overlay with closeOnSelect if open prop is set

danrot:
### Description

I have adjusted the logic happening when the `onCloseSelect` prop is set. It now also calls the `onClose` callback if the `open` prop is set to some value.

### Motivation and Context

In v2 of this library we were using this component somehow like this (simplified example):

```javascript
<ReactDateTime
    closeOnSelect={true}
    onClose={this.handleClose}
    open={this.props.open}
/>
```

This doesn't seem to work anymore in v3, because the `closeOnSelect` prop does not have any effect if any value other than `undefined` is passed to the `open` prop.

If this PR is accepted, the `handleClose` method will be called as expected. It is still in control of the rendering component if it changes the value passed to the `open` prop, in case that was a concern.

### Checklist

```
[ ] I have not included any built dist files (us maintainers do that prior to a new release)
```


danrot:
The checklist says I should not include any built dist files, but if I don't the tests and therefore the precommit hook fail :confused:

danrot:
@arqex Any chance that this is merged? This is a breaking change, that would force me to rewrite quite a bit of my application logic.

onlined:
@danrot You can use `onChange` prop for your purpose, like this:
```js
<ReactDateTime
    onChange={this.handleClose}
	{...otherProps}
/>
```

danrot:
@onlined I know that I could rewrite it this way, that's what I meant in my previous comment :slightly_smiling_face: I would just like to avoid it, and I still think that the behavior "fixed" in this PR is very confusing, and it was working differently in a previous version. In addition to that I cannot see how anybody would benefit from that change, so I guess it was on accident.

alexander-schranz:
@arqex any update to this?

alexander-schranz:
@onlined the problem with the `onChange` is that it also close automatically then when the `time` is changed and thats a behaviour you don't want e.g.:

datetime-picker

PR #769, opened 2021-01-08
Support allowSyntheticDefaultImports: false

bsl-zcs:
Now that ES6-ported class exported properly as default, typings still forcing us to import it like "`* as DatetimeModule`" and then extract its default field to use as a component:
`const Datetime = (DatetimeModule as any).default;`


bsl-zcs:
i guess trying to edit it directly on github was a bad idea. sorry.

PR #779, opened 2021-03-01
[Feature] Add useFixedPosition prop to render picker outside an "overflow: hidden" div

AdrielD:
### Description
This PR adds new optional flag, so users can force fixed position rendering and then calculating it's position based on datepicker's wrapper. This is useful in cases where the datepicker is rendered inside a div with hidden overflow and the view is being cropped.

### Motivation and Context
I developed this solution while working on a dynamic table that would scroll laterally, with datepickers as headers. When the table had just a few or no rows, the div container would severely crop datepicker's view.

Default datepicker
datepicker1

WIth applied solution:
datepicker2

Also, this PR should solve [issue 186](https://github.com/arqex/react-datetime/issues/186)

### Checklist
```
[X] I have not included any built dist files (us maintainers do that prior to a new release)
[X] I have added tests covering my changes
[X] All new and existing tests pass
[X] My changes required the documentation to be updated
  [X] I have updated the documentation accordingly
  [X] I have updated the TypeScript 1.8 type definitions accordingly
  [X] I have updated the TypeScript 2.0+ type definitions accordingly
```


erpardeepjain:
This should be merged, much need PR

TonniDev:
+1 to merge this.

PR #791, opened 2021-05-11
Add position prop

rohit267:
### Description

Added `position` prop to show the datepicker above the input feild. It accepts `up` as value
### Motivation and Context

This should be helpful when the input field is at the bottom of the screen. In this case the datepicker is overflowing the sreen without any scroll. This prop solves the proplem.
### Checklist

```
[ *] I have not included any built dist files (us maintainers do that prior to a new release)
[ *] I have added tests covering my changes
[* ] All new and existing tests pass
[* ] My changes required the documentation to be updated
  [ *] I have updated the documentation accordingly
```


rohit267:
@arqex please consider this

css/react-datetime.css
```diff
@@ -16,6 +16,10 @@
   box-shadow: 0 1px 3px rgba(0,0,0,.1);
   border: 1px solid #f9f9f9;
 }
+.rdtUp{
+  position: absolute;
+  top: -306px;
+}
```
IndrekV, line 22: Adding  these 2 styles to `rdtPicker` would fix the open in top problem reliably. Meaning that the popup will always be in the correct place no matter if in calendar or time view which have different heights. The static -306px top position would only work for one of the views.

```
.rdtPicker {
  width: 100%;
  bottom: 100%;
}
```

rohit267: ok maybe this can solve the problem of going out of the screen, if anyone want the datepicker above the input field, this prop can be helpful, a little refraction may be needed.

PR #803, opened 2021-09-06
Call correct log method in DateTime component to prevent crash on invalid input

nnatter:
### Description

This PR adjusts the code of the `DateTime` component to call the correct log method. At the moment, this code crashes the component if the `value` prop of the `DateTime` component does not match the expected format. See https://github.com/arqex/react-datetime/issues/801

### Motivation and Context

The `log` method was moved outside the class in https://github.com/arqex/react-datetime/commit/d4450038cd901f1e0f31bcd365448820f60c6f67#diff-cae892f0886b91915576f17fd54c818f7ef44b721a3885d4a742039f1c8cad03L557-R611
Unfortunately, the call this method inside of `setViewDate` was not adjusted. This causes the error described above.

### Checklist

```
```

### DIFF
--- a/src/DateTime.js
+++ b/src/DateTime.js
@@ -531,9 +531,8 @@ export default class Datetime extends React.Component {
 	 * @public
 	 */
 	setViewDate( date ) {
-		let me = this;
 		let logError = function() {
-			return me.log( 'Invalid date passed to the `setViewDate` method: ' + date );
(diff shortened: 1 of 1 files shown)


arqex:
Thanks @nnatter. It was a small refactor that was being made (and not finished), and somehow it got in to the master branch when merging the changes for the last version.

PR #813, opened 2021-11-26
Add a "browser" field so build tools can find UMD build

davidwallacejackson:
### Description
Right now, the `package.json` only indicates the presence of the CommonJS build of `react-datetime` (as `"main"`). This change exposes the UMD build via the `"browser"` field.

### Motivation and Context
Right now, Rollup loads the CJS version of `react-datetime` through [@rollup/plugin-commonjs](https://github.com/rollup/plugins/tree/master/packages/commonjs). That should work just fine, but there are some edge cases in the plugin -- [this one](https://github.com/rollup/plugins/issues/1046) in particular seems problematic for `react-datetime`.

I discovered all this by trying to import `react-datetime` in Vite and discovering that the library worked in development (where Vite implements its own esbuild-based CJS -> ESM transformation) but not in production (where Vite uses Rollup). Here's a [minimal repro](https://github.com/davidwallacejackson/repro/tree/cjs-vite).

Including ESM directly would be ideal, but webpack still lists ESM output support as "experimental" in their docs, and I'm assuming this project doesn't particularly want to migrate to another build tool. I tried this out in my local copy and it fixed the issue.

### Checklist

```
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "url": "https://github.com/arqex/react-datetime"
   },
   "main": "./dist/react-datetime.cjs.js",
+  "browser": "./dist/react-datetime.umd.js",
   "typings": "./typings/DateTime.d.ts",
   "files": [
     "css",


morlay:
the `umd` could be the `main`.

davidwallacejackson:
@morlay it could -- I'm not sure of the merits/appropriateness of using umd as the main entrypoint? I did this with `"browser"` because I found other projects doing it, and it fixed the issue I was having in Vite, but I have no strong reason to prefer this approach over the other.

fabianonunes:
It may not be necessary to export the library in another format. I think passing [libraryExport](https://v4.webpack.js.org/configuration/output/#outputlibraryexport)="default" to webpack config is enough:

```diff
// config/webpack.config.build.js
const cjsConfig = {
	...baseConfig,
	output: {
		path: outputPath,
		library: 'Datetime',
		libraryTarget: 'commonjs2',
		filename: 'react-datetime.cjs.js',
		auxiliaryComment: 'React datetime',
+               libraryExport: "default"
	}
};
```

This isn't a Vite issue but a [@rollup/plugin-commonjs](https://github.com/rollup/plugins/tree/master/packages/commonjs) bug introduced by the refactoring done in rollup/plugins#537. To confirm, you can downgrade `@rollup/plugin-commonjs` to version 15.0.0.

